PR #106, opened 2015-03-12
Fixes custom config passing to logging_configuration

tvanhens:
Timbre set-config! uses assoc-in to update the taoensso.timbre/config logging configuration. As is, this causes the timbre config map to merge an extra key {nil <new-config>} effectively negating the configuration changes. You can get around this by passing in nil merging the key {nil nil} to have onyx ignore logging configuration but this does not seem like the intended behavior.

### DIFF
--- a/src/onyx/logging_configuration.clj
+++ b/src/onyx/logging_configuration.clj
@@ -8,7 +8,7 @@
 
   (start [component]
     (if config
-      (timbre/set-config! [] config)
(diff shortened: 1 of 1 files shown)


MichaelDrogalis:
Thanks for the fix @tvanhens. Can you change the base branch to `master`? We'll put this out with the next `0.6.0` alpha.

PR #120, opened 2015-03-16
function params could not reference injected lifecycle resources

CmdrDats:
function params could not reference injected lifecycle resources which is a bit prohibitive if you want to, for instance, send the async channel along into the function call..


CmdrDats:
btw - please check that this is actually a good idea, since this kinda ties parameterization to work like this forever. It is how I would have expected the params to be working - I assumed that's the reason you'd specify the task-map keys in the params instead of the direct values.

MichaelDrogalis:
This is intentional. Use other methods of parameterization (e.g. `onyx.core/params`) to accomplish this.

PR #126, opened 2015-03-22
Allocation of site resources as part of the log

lbradstreet:
This PR isn't complete. I've created it for the purpose of discussion for now. There's definitely a race condition of some kind, which is why there's a sleep in there for now. I'll figure it out :).

The main thing I'd like a review on, is that peers stitching themselves in now use :accept-join-cluster like peers stitched in by others. I'm not sure if this is a good idea - I'm guessing you specifically avoided this originally. This simplifies the code somewhat, as it avoids code from being repeated - almost everything a peer needs to do on join is now part of :accept-join-cluster, whether self-stitched or not.

The initial peer-link should probably be moved out of peer-link as there is the possibility of a race condition now that we're setting up all of the publications in one go.


lbradstreet:
I just had the realisation that the instant-join method you used might be to avoid precisely the race condition I'm seeing.

MichaelDrogalis:
Cool! Glad you found it. :) Happy to explain if there are any particular questions.

lbradstreet:
Well I haven't tested it, so just to check, that's why you did the initial stitch in that way right?

MichaelDrogalis:
Yeah. I can't recall the exact reason, but I knew it had to be a 3-phase join process. Otherwise a peer could mistakenly pick to stitch with that it shouldn't have.

lbradstreet:
Assuming the initial two peers need to be stitched in without any coordination, I think this will complicate the assignment of peer site resources (e.g. ports). Something for me to think about tomorrow.

MichaelDrogalis:
Hm - well there is coordination, of course, in that all messages get sequentially processed.

src/onyx/messaging/aeron.clj
```diff
@@ -17,6 +17,38 @@
...
+      (assoc component :media-driver media-driver)))
+
+  (stop [{:keys [media-driver] :as component}]
+    (.close media-driver)
+    (assoc component :media-driver nil)))
+
+(defn aeron-peer-group [opts]
+  (map->AeronPeerGroup {:opts opts}))
+
+(defmethod extensions/assign-site-resources :aeron
```
lbradstreet: @MichaelDrogalis dispatch via keyword here means we don't have to actually setup a messaging component when playing back the log (which would also mean we'd need the config). I could see forgetting about it and trying to dispatch on type though.

src/onyx/messaging/aeron.clj
```diff
@@ -17,6 +17,38 @@
...
+  (let [start-port (:aeron/start-port config)
+        existing-ports (->> (vals peer-sites)
+                            (filter 
+                              (fn [s]
+                                (and (= (:aeron/bind-addr peer-site) 
+                                        (:aeron/bind-addr s))
+                                     (= (:aeron/external-addr peer-site) 
+                                        (:aeron/external-addr s)))))
+                            (map :aeron/port)
+                            set)
```
lbradstreet: @MichaelDrogalis I *think* allocating by unique bind-addr/external-addr combinations makes sense. If you have a different external-addr to your bind-addr it's probably safe to use the same port because they might be just two different containers which use the same local address.

test/onyx/log/full_test.clj
```diff
@@ -15,29 +16,33 @@
...
+                             (if (< (count (:pairs new-replica)) n-peers)
+                               (recur new-replica)
+                               new-replica)))]
+             (fact (:prepared replica) => {})
+             (fact (:accepted replica) => {})
+             (fact (set (keys (:pairs replica)))
+                   => (set (vals (:pairs replica))))
+             (fact (count (:peers replica)) => n-peers))
+
+           (finally
```
lbradstreet: @MichaelDrogalis I was a bit sick of non-shutdown envs. How's this approach for now?

src/onyx/messaging/common.clj
```diff
@@ -3,6 +3,44 @@
 (defn my-remote-ip []
   (apply str (butlast (slurp "http://checkip.amazonaws.com"))))
 
-(defn choose-ip [opts]
-  (or (:onyx.messaging/ip opts) (my-remote-ip)))
+(defn bind-addr [peer-messaging-config]
+  (:peer/bind-addr peer-messaging-config))
+
```
lbradstreet: @MichaelDrogalis code below should probably go somewhere else.

src/onyx/messaging/common.clj
```diff
@@ -3,6 +3,44 @@
...
+(defn external-addr [peer-messaging-config]
+  (or (:peer/external-addr peer-messaging-config)
+      (bind-addr peer-messaging-config)))
+
```
lbradstreet: @MichaelDrogalis code below should probably go somewhere else.

resources/test-config.edn
```diff
@@ -6,8 +6,9 @@
  :peer-config
  {:zookeeper/address "127.0.0.1:2188"
   :onyx.peer/job-scheduler :onyx.job-scheduler/greedy
-  :onyx.messaging/impl :aeron
-  :onyx.messaging/ip "localhost"}
+  :onyx.messaging/config {:messaging/impl :aeron
```
MichaelDrogalis: I'd prefer to keep the configuration flat for two reasons. 1 - namespacing at the innermost level (`messaging/impl`) starts to lose the essence of why it's namespaced to begin with. 2 - flat configurations are easier to drop into databases. They require less mangling for the user. Is there something particularly annoying about flat configurations in your usage?

src/onyx/log/zookeeper.clj
```diff
@@ -151,15 +155,31 @@
       (do (Thread/sleep 500)
           (recur)))))
 
+(defn find-messaging [log]
```
MichaelDrogalis: We should come up with a way to abort the subscription if this, or the job scheduler, can't be found and the user doesn't want to wait.

src/onyx/messaging/aeron.clj
```diff
@@ -17,6 +17,38 @@
              [java.util.function Consumer]
              [java.util.concurrent TimeUnit]))
 
+(def max-port 50000)
```
MichaelDrogalis: Let's pull this in as a configuration option.

src/onyx/messaging/aeron.clj
MichaelDrogalis: Yes, and the nice part about this implementation is that the port range is reliability chosen - meaning AWS ports can be opened for a certain number of peers and that should match up every time.

src/onyx/messaging/aeron.clj
```diff
@@ -72,99 +103,99 @@
...
 
 (defn aeron [opts]
   (map->AeronConnection {:opts opts}))
 
-(defmethod extensions/send-peer-site AeronConnection
+(defmethod extensions/peer-site AeronConnection
   [messenger]
-  {:channel (:channel messenger)
-   :stream-id (:send-stream-id messenger)})
+  {:aeron/bind-addr (:bind-addr messenger)
```
MichaelDrogalis: Why do we need both bind-addr and remote-addr?

src/onyx/messaging/aeron.clj
```diff
@@ -181,25 +212,28 @@
...
         offer-f (fn [] (.offer pub unsafe-buffer 0 len))]
     (while (not (offer-f)))))
 
 (defmethod extensions/internal-ack-message AeronConnection
   [messenger event peer-link message-id completion-id ack-val]
   (let [unsafe-buffer (protocol/build-acker-message message-id completion-id ack-val)
-        pub ^uk.co.real_logic.aeron.Publication (:pub peer-link)
+        pub ^uk.co.real_logic.aeron.Publication (:acker-pub peer-link)
         offer-f (fn [] (.offer pub unsafe-buffer 0 protocol/ack-msg-length))]
     (while (not (offer-f)))))
```
MichaelDrogalis: This might be a good place to make use of the new to-be-implemented `retry` function on an input, rather than hammering away at backpressure.

src/onyx/messaging/common.clj
MichaelDrogalis: I'm okay with that.

src/onyx/messaging/common.clj
```diff
@@ -3,6 +3,44 @@
...
+(defmethod messaging-require :aeron [_]
+  (require 'onyx.messaging.aeron))
+
+(defmethod messenger :aeron [_]
+  (ns-resolve 'onyx.messaging.aeron 'aeron))
+
+(defmethod messaging-peer-group :aeron [_]
+  (ns-resolve 'onyx.messaging.aeron 'aeron-peer-group))
+
+(defmethod messaging-require :http-kit-websockets [_]
```
MichaelDrogalis: We can rip out websockets in this PR if you want.

src/onyx/messaging/aeron.clj
lbradstreet: Without both, the port assignment gets messy for containers and it must be reproducible by the other peers. 

Imagine two containers on the same physical host:
A internal 192.168.1.1 external 10.0.0.1
B internal 192.168.1.2 external 10.0.0.1

Without both external and internal IPs, port allocation will be A 40200, B, 40201. This will be annoying and non deterministic when forwarding ports from the physical host to the container.

After writing it, I now realise the flaw with this approach is that the advertised port will be different to the forwarded port.

Maybe we need some kind of per peer advertised min port / max port too. Will have to think about it some more. For now we can just use the external ip so ports map directly between the hosts and the containers.

resources/test-config.edn
lbradstreet: No problem. I was just trying to distinguish between machine config options and cluster options, but you're right.

src/onyx/messaging/aeron.clj
lbradstreet: I meant to come back to that. Though honestly it's only really used for the range. I mean we die if we run out of ports anyway, right? It may be better to just pick a maximum number of peers we assume can ever run and add it on to the start port? If we ever hit that number we abort allocation - it's probably because peers aren't ckeaning up properly anyway?

src/onyx/messaging/aeron.clj
MichaelDrogalis: Makes sense!

src/onyx/messaging/aeron.clj
MichaelDrogalis: That's a good point - and most machines won't honestly be able to run more than a few dozen peers before having CPU lock up. Let's leave it as is.

PR #150, opened 2015-04-20
Fix/peer tests

MichaelDrogalis:
Using this to track the diff.


lbradstreet:
I think this one is ready to merge to master after a review from @MichaelDrogalis :)

PR #154, opened 2015-04-23
Core async messaging

MichaelDrogalis:
@lbradstreet Ready for review and merge. It's *fast*. Nice idea. :)


MichaelDrogalis:
Looks like there might be a problem here or there. Build is failing in some interesting ways - sometimes even in Netty. Problem may have already been present.

lbradstreet:
I've been working on some of the build failures. Messaging looks good though I think we should consider whether we can have a default receive-messages implementation - so far it's just been copy paste.

PR #187, opened 2015-05-13
Doc fixes

gardnervickers:
These are some doc fixes that have caused me minor trouble while learning Onyx.


MichaelDrogalis:
Thanks! I'll fix the Table of Contents before the release, so no worries about changing those manually.

PR #194, opened 2015-05-17
Intelligently pick ackers across physical machines.

MichaelDrogalis:
Sorts acker candidates, putting peers that are executing exempt tasks on the same machine latest in the candidate pool. cc @lbradstreet


lbradstreet:
Looks good. Will be a little while before I get a chance to bench it

PR #204, opened 2015-05-24
Pct grouping

### LINKED ISSUE #202: Implement grouping for percentage scheduler
I didn't get around to this in the percentage scheduler rewrite

MichaelDrogalis:
Fixes #202. @lbradstreet


lbradstreet:
After reading the code I'm not clear on how the initial allocation operates. Maybe you could explain whether it does what I think.

I think if you have 3 tasks:
A: min-peers 3, 50%
B: 25%, max-peers 1
C: 25%

With 10 peers, I think the initial allocation should be:
A 6
B 1
C 3

i.e. if min-peers can't be achieved the job shouldn't start, otherwise it should be ignored. As for max-peers, if it'd be allocated anyway by percentage, then take their peers out of the pool and start again.

Ah, I think the reason why I'm confused is I don't see any use of max-peers in the percentage_task_scheduler code.

MichaelDrogalis:
I started to try and read the code to see if that would work, but instead I wrote a generate test and added it to the patch. Nominally, it turns up with the configuration that you described. Occasionally, 3 and 6 are swapped. I *think* this is due to the order that peers join and the fact that grouping is sticky, but I'll investigate further.

MichaelDrogalis:
I worked the same test backwards from the replica log. We see alternate behavior when job submission does not come last. If 7 peers join, *then* the job is submitted, *then* 3 more peers join, the breakdown will be 1 / 4 / 5. The 4 peers will remain sticky, and the rest of the peers get allocated to the last task.

Keep in mind that only grouped tasks can have min-peers. It doesn't make much sense IMO to allow that knob for non-stateful tasks.

In short, I think we're okay here. But feel free to add some more tests and bang on it.

PR #206, opened 2015-05-24
Curator

lbradstreet:
Not so sure about https://github.com/onyx-platform/onyx/compare/master...curator?expand=1#diff-f93ca555f6dff62c1836d6594b9cd0b4R217 . We might be better off stopping the subscriber cleanly instead. I think this is pretty equivalent to what's going on with the current zk client.

I'm happy to change the curator namespace alias later if you like. It's currently convenient for switching back and forth.


MichaelDrogalis:
To your first comment: Yes, catching and closing on IllegalStateException wasn't my best idea. Something is probably quite wrong if that's thrown.

MichaelDrogalis:
I don't understand your note about the namespace alias, but it sounds like if I did, I'd want that changed before it's in master ;)

This patch looks great! Ship it. Big step forward for stability.

Issue #208, opened 2015-05-25
GC restart bug


MichaelDrogalis:
If GC happens when a log subscriber is reading and it misses an entry, the subscriber needs to be able to continue without crashing.

MichaelDrogalis:
Pull request opened.

MichaelDrogalis:
Merged into master.

PR #211, opened 2015-05-27
Fix/gc ch

MichaelDrogalis:
Fix for #208. cc @lbradstreet


lbradstreet:
Looks good to me. New approach seems great.

PR #212, opened 2015-05-27
Fixes #210

### LINKED ISSUE #210: ack-daemon-timeout should be the max of all input task pending-timeouts
By setting ack-daemon-timout to the maximum of all input task timeouts, you ensure that a message will be retried before it is cleared (or at least very close to, given some clock skew)

MichaelDrogalis:


lbradstreet:
Seems reasonable, I had a feeling that could be a problem but I thought we could make it work. This is good enough anyway :)

PR #217, opened 2015-06-01
Fix the link to Dire in the docs

leathekd:
I suspect this was caused by a find/replace when Onyx moved to
onyx-platform on Github.


MichaelDrogalis:
Very good catch, thanks!

PR #222, opened 2015-06-08
fix docs link

nathants:
this is already fixed in master, but github and onyxplatform/docs default to 0.6.x, which still has a broken link.

### DIFF
--- a/doc/user-guide/what-does-it-offer.md
+++ b/doc/user-guide/what-does-it-offer.md
@@ -50,11 +50,10 @@ Onyx uses the notion of a *sentinel value* to transparently switch between strea
 
 #### Aspect Orientation
 
-Clojure functions again serve as a huge win. [Dire](https://github.com/onyx-platform/dire) is a library that supports aspects, meaning you can keep your application logic airtight away from logging, preconditions, and error handling.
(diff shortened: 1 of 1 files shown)


MichaelDrogalis:
We're releasing tomorrow and merging master into `0.6.x`, so we'll decline this patch. Typically we only take pull requests in master (latest), and optionally back port patches into previous release branches. Thank you though!

cc @lbradstreet - it looks like the build failed because of Flight Recorder again? The test suite passed though. Strange.

MichaelDrogalis:
@nathants Also - if you have a suggestion for how we can better manage our branching strategy, we'd be open to discussion. One of the downsides of this approach is that trivial things like documentation fixes sit on master, while they remain broken on the latest release branch. Perhaps we should be less strict about what goes into release branches when.

nathants:
sounds good, i opened this pr against 0.6.x and not master because that's where the break is. i did read contributing ;)

i'm assuming this will be resolved magically since it's already fixed in master. just wanted to point it out in case you weren't aware.

nathants:
@MichaelDrogalis in terms of branching strategy, i suppose you could allow doc updates like this to merge straight in, or maybe have github and onyxplatform/docs default to master instead of the latest release? probably not worth complecting the branching workflow for docs though.

MichaelDrogalis:
@nathants Meh, yeah it's kinda dumb to hold up obviously broken things that don't affect code correctness. I'll merge this - seems like a good thing to do going forward. Thanks for the suggestion.

nathants:
true-flexibility++

PR #227, opened 2015-06-26
Pipeline Extensions implementation via protocols

lbradstreet:
Onyx plugins now require a path to a builder fn that builds that pipeline,i.e. :ns/fn 
Java plugins can be dispatched via :ident :ns.ClassName.
Java task fns can be resolved to static methods via :ns.Classname/methodName 

Also includes some performance optimisation work in task-lifecycle (mostly splitting out once only key lookups outside of the task-lifecycle loop).


circle-runner.sh
```diff
@@ -40,7 +40,8 @@ ARTIFACT_DIR=$CIRCLE_BUILD_NUM/$CIRCLE_NODE_INDEX/$BR"_"$1
 
 mkdir -p log_artifact/$ARTIFACT_DIR/
 
-lein with-profile dev,circle-ci jammin 360 midje $files |& tee log_artifact/$ARTIFACT_DIR/stderrout.log
+#lein with-profile dev,circle-ci jammin 360 midje $files |& tee log_artifact/$ARTIFACT_DIR/stderrout.log
```
MichaelDrogalis: Remove old Lein call.

src/onyx/peer/function.clj
```diff
@@ -19,10 +15,10 @@
     (operation/apply-function f params segment)
     segment))
 
-(defn filter-by-route [messages task-name]
-  (->> messages
-       (filter (fn [msg] (some #{task-name} (:flow (:routes msg)))))
-       (map #(dissoc % :routes :hash-group))))
+; (defn filter-by-route [messages task-name]
```
MichaelDrogalis: Remove dead code.

src/onyx/peer/function.clj
```diff
@@ -66,21 +62,66 @@
...
-            allocations (get (:allocations replica) job-id)]
-        (doseq [[[route hash-group] segs] groups]
-          (let [peers (get allocations (get egress-tasks route))
-                active-peers (filter #(= (get-in replica [:peer-state %]) :active) peers)
-                target (pick-peer id active-peers hash-group max-downstream-links)]
-            (when target
-              (let [link (operation/peer-link event target)]
-                (onyx.extensions/send-messages messenger event link segs)))))
-        {}))))
+(defn read-batch [{:keys [onyx.core/messenger] :as event}]
```
MichaelDrogalis: Is this a left over?

src/onyx/peer/function.clj
```diff
@@ -66,21 +62,66 @@
...
+  {:onyx.core/batch (onyx.extensions/receive-messages messenger event)})
+
+(defn write-batch
```
MichaelDrogalis: This too?

src/onyx/peer/pipeline_extensions.clj
```diff
@@ -1,44 +1,34 @@
...
-
-(defmulti ^{:added "0.6.0"} pending?
-  "Returns true if this message ID is pending."
-  (fn [event messsage-id]
-    (ident-dispatch event)))
-
-(defmulti ^{:added "0.6.0"} drained?
-  "Returns true if this input resource has been exhausted."
-  ident-dispatch)
+(defprotocol PipelineInput
```
MichaelDrogalis: I dont think I quite understand the split between PipelineInput and Pipeline since read-batch appears in the latter.

src/onyx/peer/task_lifecycle.clj
```diff
@@ -9,6 +9,7 @@
               [onyx.messaging.acking-daemon :as acker]
               [onyx.peer.pipeline-extensions :as p-ext]
               [onyx.peer.function :as function]
+              [onyx.plugin.core-async]
```
MichaelDrogalis: I think this should only be required via onyx.system.

src/onyx/peer/task_lifecycle.clj
```diff
@@ -461,6 +479,26 @@
...
+(defn build-pipeline [task-map pipeline-data]
+  (let [kw (:onyx/ident task-map)]
+    (try 
+      (if (#{:input :output} (:onyx/type task-map))
+        (let [user-ns (namespace kw)
+              user-fn (name kw)]
+          (if (and user-ns user-fn)
+            (if-let [f (ns-resolve (symbol user-ns) (symbol user-fn))]
+              (f pipeline-data)    
+              (throw (Exception.)))
```
MichaelDrogalis: I think we can throw a better exception here. :)

MichaelDrogalis:
Left some comments.

src/onyx/peer/task_lifecycle.clj
lbradstreet: True, though it is being caught and a new, better error is thrown. I'll improve it since we do return the inner exception in the ex-info now.

src/onyx/peer/task_lifecycle.clj
lbradstreet: Oops. That was just a hack to get it working before I did plugin resolution and it's no longer required.

src/onyx/peer/pipeline_extensions.clj
lbradstreet: Function only implements Pipeline. It mostly had to do with the fact that if you implement an interface in java you need to implement all methods on that interface, therefore it makes sense to split the protocols up. Now that we're generating the interfaces independently and then implementing the protocols (see interop.clj), this distinction might not be as important.

src/onyx/peer/function.clj
lbradstreet: Good one. I deleted that in one of my later branches.

src/onyx/peer/function.clj
lbradstreet: Hmm yes, though I think I will want to export it through interop so that it can be used by output plugins in java (see what I mean in write-batch below)

src/onyx/peer/function.clj
lbradstreet: See https://github.com/onyx-platform/onyx/blob/pext-protocol/src/onyx/plugin/core_async.clj#L38 https://github.com/onyx-platform/onyx/blob/pext-protocol/src/onyx/interop.clj#L7 and https://github.com/onyx-platform/onyx-benchmark/blob/master/src/java/onyx/plugin/RandomInputPlugin.java#L137. Potentially this should be done via an abstract class with an abstract method. I'm not sure if there's an easy way to create these in java though.

The only reason why we have two copies in there is because the version in Function doesn't require a lot of lookups because we can put a lot of the keys (replica, state, id, etc) in the Function record once and be done with it (i.e. it's an optimisation). I could move all the logic into one function and do the lookup outside though. I'll do that.

src/onyx/peer/task_lifecycle.clj
lbradstreet: Pushed a fix for this

lbradstreet:
I'd like to merge this PR soon. Main issue is whether we have the correct format for plugin resolution, i.e. `:onyx/ident ns/builder-fn` or `:onyx/ident java-ns-with-constructor`. 

I would like to at least change the key from ident to `:onyx/plugin` or something like that. I'd say `:onyx/plugin-fn` is fine but for the java case it's actually an ns, in which case you might want to split it to `:onyx/plugin-ns` and `:onyx/plugin-builder-fn` or something like that (I also hate these).

lbradstreet:
Todo: rename ack-message and retry-message to ack-segment and retry-segment. Also rename message-id and other uses of message to segment.

PR #229, opened 2015-06-29
grammar

noisesmith:
changes "sending a command to the log that it gets allocated for work" to "sending a command to the log requesting that it be allocated for work"

I think this more clearly states the intended meaning?


MichaelDrogalis:
Thanks!

PR #231, opened 2015-07-02
Updating to the most recent version of timbre.

rosejn:
We were having problems using Onyx with some internal libs because the latest Timbre moved the rotating logger namespace and changed some args.  This patch updates Onyx to use Timbre 4.0.2 and fixes the logging configuration to use the new parameters.


MichaelDrogalis:
Excellent! Thanks! We'll be cutting a new release in the near future. For now, Onyx `0.6.1-SNAPSHOT` on Clojars has your change in it.

PR #232, opened 2015-07-06
Fixed the summary links

gardnervickers:
The link paths were previously broken


MichaelDrogalis:
Links are correct and in that manner for GitBook. Thanks for checking though!

PR #233, opened 2015-07-07
Fixes #118

### LINKED ISSUE #118: Extend group-by-key in the catalog entry for composites
I see a good use case for wanting to group by multiple keys, as in `:onyx/group-by-key [:key-1 :key-2]`

gardnervickers:
Allow grouping on multiple keys.

### DIFF
--- a/src/onyx/peer/task_lifecycle.clj
+++ b/src/onyx/peer/task_lifecycle.clj
@@ -94,7 +94,9 @@
 (defn group-message [segment catalog task]
   (let [t (find-task-fast catalog task)]
     (if-let [k (:onyx/group-by-key t)]
-      (hash (get segment k))
+      (if (vector? k)
+        (hash (sort ((apply juxt k) segment)))
+        (hash (get segment k)))
       (when-let [f (:onyx/group-by-fn t)]
         (hash ((operation/resolve-fn {:onyx/fn f}) segment))))))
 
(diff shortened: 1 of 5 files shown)


gardnervickers:
I'll fix the docs too in a couple minutes

src/onyx/peer/task_lifecycle.clj
```diff
@@ -94,7 +94,9 @@
 (defn group-message [segment catalog task]
   (let [t (find-task-fast catalog task)]
     (if-let [k (:onyx/group-by-key t)]
-      (hash (get segment k))
+      (if (vector? k)
```
lbradstreet: Should probably be (sequential? k) so users can use lists and vectors.

src/onyx/peer/task_lifecycle.clj
```diff
@@ -94,7 +94,9 @@
...
+        (hash (sort ((apply juxt k) segment)))
```
lbradstreet: This use of juxt will break groupings with non-keyword keys (which is why get was used rather than (k segment)). It'd be nice to keep this property when using multiple keys.

Is there any reason why we're sorting here? The key order will be consistent over the peers, so I think grouping will be consistent without the sort.

lbradstreet:
Nice PR @gardnervickers, I left a couple of inline comments above. Cheers!

src/onyx/peer/task_lifecycle.clj
gardnervickers: Good catch. Will fix

MichaelDrogalis:
Feeling ready to merge this one, @gardnervickers?

gardnervickers:
Yea we're good to go

MichaelDrogalis:
Thanks!

PR #234, opened 2015-07-08
Onyx Internal Monitoring

MichaelDrogalis:
This feature isn't complete - but follows the general guideline of how I'd like to see monitoring of Onyx itself. This patch adds a `:monitoring` dependency to all components, and invokes callbacks on all latency-sensitive and I/O calls. The patch as of now captures all sensitive ZooKeeper activity. The API is fully backwards-compatible.


MichaelDrogalis:
@lbradstreet This is ready for a review and merge.

lbradstreet:
I made a few changes, let me know if you disagree with them.

I created a new emit-latency which takes away a lot of the cost of the calls. Before there was a penalty of 5-7%. Now it's lost in the noise, which will be important if we start adding calls in more places. I'd actually like to make emit-latency a macro so we can get away from passing in a function.

I made some record types to make for faster emit dispatch, where required.

Some of these calls are generally much too fast for ms measurements. Should we measure via nanoTime (rather that currentTimeMillis)? Perhaps most of these calls are only relevant if they start creeping into the multi millisecond range.

What's the purpose of config? As far as I can see, there isn't really a good way to get unknown keys into the record - they end up in config. Is there any reason they can't be assoc'd into the record?

Otherwise looks good. I'll still need a way to monitor the buffers later, but this could probably be achieved  by adding another call.

MichaelDrogalis:
Landed this directly into develop via command line. Closing.

PR #235, opened 2015-07-09
Expand post-transform function arity to include segment

### LINKED ISSUE #225: Allow post-transform to access the segment that caused the exception
Add another arity to this function to get the original segment.

MichaelDrogalis:
Fixes #225. @lbradstreet


lbradstreet:
Looks good!

PR #239, opened 2015-07-09
Implement retry-message and ack-message lifecycle methods

lbradstreet:
There's also a couple log command changes in there. One to fix a bug in the generative tests that were found while I was developing this. Apologies for slipping it in.


src/onyx/log/commands/leave_cluster.clj
```diff
@@ -48,13 +48,11 @@
 
 (defmethod extensions/reactions :leave-cluster
   [{:keys [args]} old new diff state]
-  (let [allocation (common/peer->allocated-job (:allocations old) (:id state))
-        scheduler (get-in new [:task-schedulers (:job allocation)])]
```
lbradstreet: allocation / scheduler were bound but unused, so I cleaned this up

src/onyx/log/commands/accept_join_cluster.clj
```diff
@@ -12,9 +12,10 @@
   (let [{:keys [accepted-joiner accepted-observer]} args
         target (or (get-in replica [:pairs accepted-observer])
                    accepted-observer)
+        accepted? (get-in replica [:accepted accepted-observer])
         already-joined? (some #{accepted-joiner} (:peers replica))
         no-observer? (not (some #{target} (:peers replica)))]
-    (if (or already-joined? no-observer?) 
+    (if (or already-joined? no-observer? (not accepted?))
```
lbradstreet: Generative tests found another bug where the peer had already left (via leave-cluster) and then accept-join-cluster came in. This was an easy fix.

lbradstreet:
Merging as the operation of lifecycle-test on docker should be improved by my latest aeron improvements

PR #244, opened 2015-07-11
Aeron share subscription

lbradstreet:
This PR allows all virtual peers to multiplex over a single subscription, which will scale as the number of peers/vpeers increases. It also allows Aeron messaging to short circuit 

The main problem that I can see with it that the subscriber thread is currently responsible for all the deserialization for all the vpeers using the same peer group. For serialization heavy operations this thread could end up being overwhelmed.

We will likely have to push the buffers onto our channels and serialize them outside of the subscriber thread. By the way, I think our netty implementation may also have this issue.

Also worth noting, impl detail wise: every message sent via aeron now includes a short which maps between the id and a virtual peer. This is necessary because we now share a subscription, and we need to know which vpeer to send the message on to. By using the short -> vpeer mapping scheme, we save 14 bytes in every message. This is allocated in the same way that ports were before. Also worth noting: aeron only needs one port per peer group now (since we multiplex).


lbradstreet:
Aeron now spins up `onyx.messaging.aeron/subscriber-count` subscribers when the peer-group starts up. Individual nodes will only create a publication to one subscriber, which is chosen at random (but is consistent). This ensures that the number of inter-node links will still scale.

lbradstreet:
@MichaelDrogalis, I think this one is fully baked now. Review and merge at your leisure

MichaelDrogalis:
@lbradstreet If you could document somewhere in the user guide what short circuiting is, that would be great. There are a few small things I want to clean up like commented out code, but I'll patch those over on develop. Great patch, thanks!

lbradstreet:
Oops. Forgot to document short circuiting! Will do.

lbradstreet:
Short circuiting document and I removed the commented out code.

PR #252, opened 2015-07-18
Restart pred fn

### LINKED ISSUE #249: :onyx/restartable-fn
In lieu of a static list of restartable-exceptions (see https://github.com/onyx-platform/onyx/blob/master/src/onyx/peer/task_lifecycle.clj#L20), I think it might be worth parameterizing a function that checks whether the exception returned by a task is restartable.

For example, I have a case where I want to test fault tolerance of the datomic plugin and would love to return a clojure.lang.ExceptionInfo with `{:restartable? true}`, and it would be trivial to add a handler for this if it was parameterizable in the input task. I'm thinking something like:

`:onyx/restartable-fn :onyx.plugin.datomic-input-test/restartable?`
where 
```clojure
(defn restartable? [e]
   (and (= (type e) clojure.lang.ExceptionInfo)
           (:restartable? (ex-info e)))
```

Plugin authors could supply catalog entries that include a restartable? fn that allows for vpeers to be restarted under allowable conditions.

cc/@MichaelDrogalis

lbradstreet:
Closes #249 
cc/@michaeldrogalis

### DIFF
--- a/doc/user-guide/information-model.md
+++ b/doc/user-guide/information-model.md
@@ -43,11 +43,13 @@ This section specifies what a valid catalog, workflow, and flow conditions look 
 
 ##### All maps may optionally have these keys
 
-| key name             | type       | choices              | default    |    Meaning                                                            |
-|----------------------|------------|----------------------|------------|-----------------------------------------------------------------------|
-|`:onyx/batch-timeout` | `integer`  | `>= 0`               | `1000`     |                                                                       |
-|`:onyx/max-peers`     | `integer`  | `> 0`                |            |                                                                       |
-|`:onyx/language`      | `keyword`  | `:clojure`, `:java`  | `:clojure` | Affects `:onyx/fn` and `:onyx/plugin` function and plugin resolution  |
+| key name              | type       | choices              | default    |    Meaning                                                            |
+|-----------------------|------------|----------------------|------------|-----------------------------------------------------------------------|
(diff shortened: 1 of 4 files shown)


doc/user-guide/information-model.md
```diff
@@ -43,11 +43,13 @@ This section specifies what a valid catalog, workflow, and flow conditions look
...
-|`:onyx/language`      | `keyword`  | `:clojure`, `:java`  | `:clojure` | Affects `:onyx/fn` and `:onyx/plugin` function and plugin resolution  |
+| key name              | type       | choices              | default    |    Meaning                                                            |
+|-----------------------|------------|----------------------|------------|-----------------------------------------------------------------------|
+|`:onyx/batch-timeout`  | `integer`  | `>= 0`               | `1000`     |                                                                       |
+|`:onyx/max-peers`      | `integer`  | `> 0`                |            |                                                                       |
+|`:onyx/language`       | `keyword`  | `:clojure`, `:java`  | `:clojure` | Affects `:onyx/fn` and `:onyx/plugin` function and plugin resolution  |
+|`:onyx/restart-pred-fn`| `keyword`  | `any`                |            | Keyword pointing to function taking an exception which returns        |
+|                       |            |                      |            | a boolean for whether a peer is restartable following that exception  |
```
MichaelDrogalis: The markdown table gets malformed here.

MichaelDrogalis: I'll fix this up on develop.

lbradstreet:
Ugh, I suck at markdown.

PR #255, opened 2015-07-20
Perf/reduce intermediate allocations

lbradstreet:
A few points to mention, it looks like a lot of extra code, but if you remove the rotating-seq code, it's actually less code.

I moved rotating-seq into onyx because I needed to make some perf changes. How do you feel about moving it into Onyx permanently?

We're now using peer-replica-view more efficiently - it is now personalised for the task the peer is running.

The weakest part of this PR, is how many different things are done in https://github.com/onyx-platform/onyx/pull/255/files#diff-00d8e37583c932cbfab7693471679346R114, but it causes a pretty big performance bump, and it actually allowed me to reduce a lot of code elsewhere.

One major change is that build-new-segments now returns a record with`:tree`, `:segments`, and `:acks`. Output plugins will need to be modified as seen here:
https://github.com/onyx-platform/onyx/pull/255/files#diff-b104639c449a180201081c8589958c7bL103

Overall perf bump with this PR is about 33% on Aeron messaging when short circuiting, and probably around 20% without.


src/onyx/extensions.clj
```diff
@@ -14,7 +14,7 @@
 
 ;; Peer replica view interface
 
-(defmulti peer-replica-view (fn [entry old new diff old-view peer-id] :default))
+(defmulti peer-replica-view (fn [entry old new diff old-view peer-id opts] :default))
```
MichaelDrogalis: I could be wrong, but isn't `peer-id` provided by `opts`?

src/onyx/log/commands/common.clj
```diff
@@ -96,7 +96,8 @@
         active? (partial at-least-one-active? replica)]
     (every? identity (map #(active? (get-in replica [:allocations job %])) tasks))))
 
-(defn transform-job-allocations [peer-state allocations]
+(defn job-receivable-peers [peer-state allocations job-id]
+  ;; can also filter down tasks?
```
MichaelDrogalis: What does this comment mean?

src/onyx/extensions.clj
lbradstreet: opts is the peer-config, and doesn't have peer-id in it

src/onyx/messaging/aeron.clj
```diff
@@ -114,30 +116,31 @@
           (.idle idle-strategy fragments-read))))))
 
 (defn handle-message [decompress-f virtual-peers buffer offset length header]
+  ;;; Problem, all de-serialization is now done in a single subscriber thread
```
MichaelDrogalis: We might want to leave a note here about how to performance tune it.

src/onyx/extensions.clj
MichaelDrogalis: Ah. *Those* opts. Got it.

src/onyx/peer/task_lifecycle.clj
```diff
@@ -69,107 +70,102 @@
...
-                (map (fn [result] (let [root (:root result)
-                                       fused-leaf-vals (gen-ack-fusion-vals task-map (:leaves result))
-                                       fused-vals (if-let [ack-val (:ack-val root)] 
-                                                    (bit-xor fused-leaf-vals ack-val)
-                                                    fused-leaf-vals)]
-                                   (->Ack (:id root)
-                                          (:completion-id root)
-                                          ;; or'ing by zero covers the case of flow conditions where an
-                                          ;; input task produces a segment that goes nowhere.
-                                          (or fused-vals 0)
```
MichaelDrogalis: Hm - trying to find where we properly handle this edge case in the patch.

src/rotating_seq/core.clj
```diff
@@ -0,0 +1,18 @@
+(ns rotating-seq.core)
```
MichaelDrogalis: Let's move this under `static/rotating_seq.clj`

src/onyx/log/commands/common.clj
lbradstreet: I'll improve it, it's mostly a TODO. What it means is that we know that a peer on a particular task should filter receivable peers further, as only peers on egress tasks are required.

src/onyx/messaging/aeron.clj
lbradstreet: Good idea

PR #266, opened 2015-07-31
Fix bulk function execution with empty segments

### LINKED ISSUE #260: Bulk functions get invoked with empty sequences
If a batch times out and returns zero segments, bulk functions are invoked anyway with the empty seq. If a batch is empty, no call should occur.

tvanhens:
Fixes #260

### DIFF
--- a/src/onyx/peer/task_lifecycle.clj
+++ b/src/onyx/peer/task_lifecycle.clj
@@ -269,7 +269,7 @@
 (defn apply-fn-bulk [f {:keys [onyx.core/batch] :as event}]
   ;; Bulk functions intentionally ignore their outputs.
   (let [segments (map :message batch)]
-    (f segments)
+    (when-not (seq segments) (f segments))
     (assoc
       event
       :onyx.core/results


MichaelDrogalis:
@tvanhens Thanks for the fix! Can we switch `empty` to `(not (seq))` per Clojure's [docs](https://clojuredocs.org/clojure.core/empty_q)? Truthfully I don't know why this is preferable, but I've seen it all over the place.

lbradstreet:
Is that actually the recommendation? I read it as use (seq x) as a form of 'not empty'

tvanhens:
`(when-not (seq segments) ...)` or `(when (not (seq segments)) ... )`
The second version feels a little gross imo.

MichaelDrogalis:
The first one is good to me.

MichaelDrogalis:
Thanks for making that (possibly frivolous) change. :)

tvanhens:
I made it an amend since I'm on a fork and its a small change. Wanted to keep the commit log clean. Hope thats alright.

MichaelDrogalis:
:+1

PR #271, opened 2015-08-17
Fixed URLs in summary page of user guide

darth10:
URLs were all 404s. Fixed it.


gardnervickers:
This is a problem with GitBook, see closed pr #232

darth10:
Ah I see! Thanks for the info. Closes #271.

PR #275, opened 2015-08-18
Add timeout to take-segments! and await-job-completion

### LINKED ISSUE #219: take-segments! and await-job-completion should take timeouts
Currently they'll wait forever for data that may never come.

gardnervickers:
fixes #219 
For `await-job-completion` I'm not sure the argument ordering. It would be nice to be able to specify quasi-builder-style `(await-job-completion peer-config job-id :monitoring-config {:monitoring :no-op} :timeout 100)` 

But for now I decided to keep it consistent with the rest of onyx.  

P.S. if anyone knows how to stop GIT from highlighting whitespace changes, I can amend this. Emacs is not loving whatever is going on here. Cant save the file without the lines getting funky.

Thanks!

### DIFF
--- a/src/onyx/api.clj
+++ b/src/onyx/api.clj
@@ -1,6 +1,6 @@
 (ns onyx.api
   (:require [clojure.string :refer [split]]
-            [clojure.core.async :refer [chan alts!! >!! <!! close!]]
+            [clojure.core.async :refer [chan alts!! >!! <!! close! alts!! timeout go]]
             [com.stuartsierra.component :as component]
             [taoensso.timbre :refer [warn fatal error]]
             [onyx.log.entry :refer [create-log-entry]]
@@ -195,9 +195,9 @@
    Local replicas clear out all data about completed and killed jobs -
    as if they never existed."
   ([peer-config]
-     (gc peer-config {:monitoring :no-op}))
+   (gc peer-config {:monitoring :no-op}))
   ([peer-config monitoring-config]
-     (let [id (java.util.UUID/randomUUID)
+   (let [id (java.util.UUID/randomUUID)
            client (component/start (system/onyx-client peer-config monitoring-config))
            entry (create-log-entry :gc {:id id})
            ch (chan 1000)]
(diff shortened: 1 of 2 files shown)


MichaelDrogalis:
I need some time to figure out a better signature for this. You can simplify a lot of the code here by using a core.async timeout channel, though. The future becomes superfluous.

gardnervickers:
is `(take-segments!)` ok?

MichaelDrogalis:
The only thing I would change is that `:done` should not appear in timed-out calls, which I believe would happen with this function.

lbradstreet:
Any thoughts on making it obvious to the user that the call timed out? The lack of :done should make it obvious but I'm putting it out there.

MichaelDrogalis:
Could return a map of status and segments, perhaps? Im okay with breaking backwards here, its just a little utility function.

lbradstreet:
I'd personally like that. I think it'd make a lot of the test failures more clear.

gardnervickers:
so
```clojure
{:status :timeout
 :segment []}

{:status :done
 :segments []}
```
Same thing for `await-job-completion`?

```clojure
{:status :done}
{:status :failed}
{:status :timeout}
```

MichaelDrogalis:
I'm less enthusiastic about this change today. We've used `take-segments!` in a *ton* of places beyond the Onyx tests. It's in the examples, workshop code, plugins, etc. I think I'm willing to settle for timing-out being indicative of `:done` not being present. I also don't sweat this much as we're planning to remove `:done` anyway. Does this seem okay to both of you?

src/onyx/api.clj
```diff
@@ -218,22 +218,28 @@
...
-                 :else 
-                 (recur new-replica)))))))
+   (await-job-completion peer-config job-id {:monitoring :no-op} nil))
+  ([peer-config job-id monitoring-config timeout-ms]
+   (-> (future
+         (let [job-id (validator/coerce-uuid job-id)
+               client (component/start (system/onyx-client peer-config monitoring-config))
+               ch (chan 100)]
+           (loop [replica (extensions/subscribe-to-log (:log client) ch)]
+             (let [entry (<!! ch)
```
MichaelDrogalis: You can do the same trick you're doing below with a timeout channel to obviate using a `future`.

gardnervickers: Yea I have 

```clojure

(defn ^{:added "0.6.0"} await-job-completion
  "Blocks until job-id has had all of its tasks completed or the job is killed.
   Returns true if the job completed successfully, false if the job was killed."
  ([peer-config job-id monitoring-config timeout-ms]
   (let [job-id (validator/coerce-uuid job-id)
         client (component/start (system/onyx-client peer-config monitoring-config))
         ch (chan 100)
         tmt (timeout timeout-ms)]
     (loop [replica (extensions/subscribe-to-log (:log client) ch)]
       (let [[v c] (alts!! [tmt (go (extensions/apply-log-entry (<!! ch)
                                                                replica))])]
         (cond (some #{job-id} (:completed-jobs v))
               (do (component/stop client)
                   true)
               (some #{job-id} (:killed-jobs v))
               (do (component/stop client)
                   false)
               (= c tmt)
               (do (component/stop client)
                   :timeout)
               :else
               (recur v)))))))
```
Just having some insane problems with getting my commits to look nice.

MichaelDrogalis: Haha okay, yeah that's the ticket. Should be ready to go when you get sorted out.

gardnervickers:
@MichaelDrogalis Yes, very good!

lbradstreet:
Sounds good to me, since we're removing the done later.

MichaelDrogalis:
That's our phantom failure test. :( Unrelated.

PR #278, opened 2015-08-20
fix minor docs error on lifecycle function params

mccraigmccraig:
keywords for setting function params from a before-task-start lifecycle call were wrong


MichaelDrogalis:
Thanks!

PR #280, opened 2015-08-25
Use the peer id of the joiner to generate the aeron id.

dignati:


dignati:
Maybe use the hash as an index in the vec of available short-ids instead? The worst case complexity is pretty bad with this change.

lbradstreet:
You're right, that's the big flaw in this approach. Yes, maybe it's better to filter the allocated ids out, then hash mod, and pull out the element with that index.

dignati:
Much better code as well.

MichaelDrogalis:
:+1:

lbradstreet:
Perfect - thanks for knocking this one off!

PR #285, opened 2015-09-03
Base replica map and stricter replica schema

lbradstreet:


MichaelDrogalis:
Good to me. Merge when it passes. :)

PR #286, opened 2015-09-03
Fix SUMMARY links.

ideal-knee:
These links are broken when viewing on github.com.  Perhaps they do the right thing when generating docs in some other way?


MichaelDrogalis:
See #232 and and #271. I wish we could put a comment in that markdown file or something. Thanks for sending this though, we appreciate the effort!

ideal-knee:
Thanks for the quick response.  How can I access the output of gitbook?

MichaelDrogalis:
Sure thing.

- [Main page](https://www.gitbook.com/book/onyx-platform/onyx/details)
- [HTTP version](http://onyx-platform.gitbooks.io/onyx/content/)

ideal-knee:
Thanks! I haven't seen those links anywhere on onyxplatform.org or from github.com/onyx-platform/onyx.  Following the documentation links from onyxplatform.org leads to those broken links, which is not great from a marketing perspective.  Perhaps if I was more Github savvy I would've found that book. :-P

MichaelDrogalis:
Thanks, I made the link on the website point to GitBook. Sometimes I forgot how differently I search the Onyx codebase and docs than other people. :)

PR #288, opened 2015-09-03
Allow task-metadata to be cleaned up from plugins

lbradstreet:


lbradstreet:
Needs a quick QC from @MichaelDrogalis if he has a moment. I'll put this one in 0.7.3.

MichaelDrogalis:
This should also be dissoc'ed during GC

lbradstreet:
Damn. I knew I forgot something.

lbradstreet:
I think this is enough. A quick re-qc after build pass would be good though.

MichaelDrogalis:
Good to go! Thanks:)

Issue #245, opened 2015-07-11
Experiment with segment write queue


lbradstreet:
Workflows with high segments branching factors (i.e. return a lot of segments) may lead to huge batches being sent to their egress tasks, even though the task may only read smaller batch sizes. It may be useful to have a write queue in between, which is read by the messaging layer and then sent in appropriate batch sizes.

Reading the Heron paper (5.5.2) it appear that they are doing something like this. Note, they experienced some GC issues with this approach which required them to dynamically adjust the queue sizes.

lbradstreet:
No longer relevant in 0.10.

PR #290, opened 2015-09-07
Aeron publication management / write queue

lbradstreet:
Adds an additional layer above the publication that will do management on Aeron publications and write buffering. This solves the issue where the Aeron connection is not up yet writes are being tried, without holding up the pipeline until the publication is up. We can thus bump our Aeron dependencies.

This doesn't solve https://github.com/onyx-platform/onyx/issues/245 yet, as it doesn't split up batches, but it's a step in the right direction.


lbradstreet:
See note here: https://github.com/onyx-platform/onyx/pull/290/files#diff-6f513ce68f2a1a15a3820e3ce9c5db42R46

Ideally this would be non-blocking, or at least send an event out to monitoring when it doesn't succeed, and then retry. Unfortunately, offer! was committed to core.async many months ago and there hasn't been a release yet.

MichaelDrogalis:
We should definitely give the core.async crew a shout, I've needed `offer!` many times. Will review soon.

MichaelDrogalis:
@lbradstreet This looks good to me. Ready for merging!

PR #299, opened 2015-09-15
Slot based grouping

lbradstreet:
Implements slot id allocation for use in grouping tasks. Peer selection has been moved out task-lifecycle and into peer-replica-view which now has more responsibility.

More invariants are now checked in the property tests


src/onyx/log/commands/kill_job.clj
```diff
@@ -21,6 +21,7 @@
           (update-in [:allocations] dissoc job-id)
           (update-in [:task-metadata] dissoc job-id)
           (update-in [:ackers] dissoc job-id)
+          (update-in [:task-slot-ids] dissoc job-id)
```
MichaelDrogalis: Task slots should also get collected during a GC.

MichaelDrogalis:
One small fix with GC, otherwise merge it! Great work!!

PR #304, opened 2015-09-19
Windows

MichaelDrogalis:
cc @lbradstreet


doc/user-guide/information-model.md
```diff
@@ -116,6 +119,57 @@ This section specifies what a valid catalog, workflow, and flow conditions look
...
+|`:window/type`        |`string`   | no       |
+|`:window/aggregation` |`string`   | no       |
+|`:window/window-key`  |`string`   | no       |
+|`:window/min-value`   |`int`      | yes      |
+|`:window/range`       |`vector`   | no       |
+|`:window/slide`       |`vector`   | sometimes|
+|`:window/init`        |`string`   | sometimes|
+|`:window/doc`         |`string`   | yes      |
+
+`:window/range` and `:window/slide` are values that require Units. See below for a description of Units.
```
lbradstreet: It'd be nice have some examples when we describe the information model, though I guess these could be left for the doc that discusses windowing/triggers.

doc/user-guide/windowing.md
```diff
@@ -0,0 +1,187 @@
...
+
+See the Information Model chapter for an exact specification of what values the Window maps need to supply. Here we will describe what each of the keys mean.
+
+| key name             |description
+|----------------------|-----------
+|`:window/id`          | A unique identifier per window
+|`:window/task`        | The workflow task over which the window operates
+|`:window/type`        | Which type of window this is (fixed, sliding, etc)
+|`:window/aggregation` | The aggregation function to apply, as described above
+|`:window/window-key`  | The key over which range
```
lbradstreet: Description could be better here "The key over which the range will be calculated?

doc/user-guide/windowing.md
```diff
@@ -0,0 +1,187 @@
...
+
+The `:sum` operation adds the values of `:window/sum-key` for all segments in the window.
+
+```clojure
+{:window/id :sum-ages
+ :window/task :identity
+ :window/type :fixed
+ :window/aggregation :sum
+ :window/sum-key :age
+ :window/window-key :event-time
```
lbradstreet: As an alternative to using :window/aggregation X, :window/X-key, how about:
`:window/aggregation [aggregation key]`, e.g. `:window/aggregation [:sum :age]`. This makes it a lot easier to switch aggregation types programatically.

doc/user-guide/windowing.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+See the information model for all supported units. You can use a singular form (e.g. `:minute`) instead of the plural (e.g. `:minutes`) where it makes sense for readability.
+
+Onyx is also capable of sliding by `:elements`. This is often referred to as "slide-by-tuple" in research. Onyx doesn't require a time-based range and slide value. Any totally ordered value will work equivalently.
+
+### Aggregation
+
+Windows allow you accrete data over time. Sometimes, you want to store all the data. Othertimes you want to incrementally compact the data. Window specifications must provide a `:window/aggregation` key. We'll go over an example of every type of aggregation that Onyx supports.
+
```
lbradstreet: Should we provide an ability to supply a `:window/aggregation-fn`?, or do we expect them to implement a multimethod. The latter may be less portable.

doc/user-guide/windowing.md
MichaelDrogalis: This is intentionally kept separate in preparation for the ability to make the range and slide attributes over two different values. It's described in the latter part of the WID paper, which I did not implement yet. So then we'd have `:window/range-key` and `:window/slide-key` when we want to vary them. Other windows will likely require different key patterns.

doc/user-guide/windowing.md
MichaelDrogalis: Ah, I didn't even think of this. Can you think of any other places where multimethods are still the defacto extension point? One downside I can see is having to specify a longer, namespaced keyword for common things like min, max, and sum. But that's a more of a creature comfort.

doc/user-guide/windowing.md
MichaelDrogalis: Fixing.

doc/user-guide/windowing.md
lbradstreet: I think this is the only one. We can always have some short form keywords that we expand out to the full ns when we resolve.

MichaelDrogalis: Cool. Thanks for bringing this up, I would have totally missed it otherwise. I got carried away with hammering out the implementation.

PR #306, opened 2015-09-25
Fixed invalid link in plugin documentation

KevinGreene:


lbradstreet:
Ah. Thanks. I actually removed that log command completely now that we now do it in onyx internally. I'll just delete that sentence instead. Cheers!

KevinGreene:
My mistake, didn't fully look at the surrounding documentation. I'm not 100% sure, but I think the paragraph before, and maybe a little more, should be deleted, but I don't know for sure.

MichaelDrogalis:
@lbradstreet Closing, unless you wanted anything else from this patch. Thanks @KevinGreene!

lbradstreet:
Thanks @KevinGreene, cleaned up in 669a226

PR #326, opened 2015-10-04
Fixes #310.

### LINKED ISSUE #310: Add monitoring hook for network bytes written
This should have some way of tracking back to the task that created them (task-id), and should be called after the point of serialisation

MichaelDrogalis:


lbradstreet:
As a general rule, any monitoring in the critical path should use event records. Rather than placing task metadata in the events themselves, I think it's preferable to assoc these into the monitoring record for use in any calls that might be made. This saves us a map lookup and ensures relevant task data is always there.

Happy to merge and close now if you agree.

MichaelDrogalis:
Great call. Merging.

PR #327, opened 2015-10-06
Fixes #134.

### LINKED ISSUE #134: Tear out "immediate" property of log entries
As discussed, this design decision doesn't actually defend against anything.

MichaelDrogalis:
Fixes #134. We should review this one very carefully and make sure this is still the right decision. @lbradstreet


MichaelDrogalis:
Reviewed. We're good here. The reason I added a buffer to hold back some messages was due to the way the scheduler was initially designed - e.g. "volunteering for tasks" on an entry by entry basis.

PR #330, opened 2015-10-08
State Management

lbradstreet:
Includes a full implementation of state management implemented via BookKeeper.
Acking is achieved for async bookkeeper writes, by incrementing a reference count on the ack for every action that needs to be completed before an ack is sent. When each action is completed, this counter is decremented, and the action that finally reduces this count to zero is the action that will send the ack out.

Unique segment tracking is performed by setting a key :onyx/uniqueness-key which is set on the task-map. Currently this uniquness tracking is performed via a big set that is never flushed. This will be solved in a future PR.


lbradstreet:
A few items of note. We now supported both batched acks as well as single acks. Batched acks are generally used unless an ack's reference count is reduced to zero somewhere other than the ack-segments call in task-lifecycle.

Documentation is currently lacking. This PR should not be merged until I have added it.

lbradstreet:
:onyx.peer/state-filter-impl should probably be selectable via task-map options. There are probably cases where a bloom filter is good enough, without any secondary lookup.

Thoughts?

src/onyx/schema.clj
```diff
@@ -185,12 +195,21 @@
...
-   (s/optional-key :onyx.peer/peer-not-ready-back-off) s/Int
    (s/optional-key :onyx.peer/job-not-ready-back-off) s/Int
+   (s/optional-key :onyx.peer/peer-not-ready-back-off) s/Int
    (s/optional-key :onyx.peer/fn-params) s/Any
    (s/optional-key :onyx.peer/backpressure-check-interval) s/Int
    (s/optional-key :onyx.peer/backpressure-low-water-pct) s/Int
    (s/optional-key :onyx.peer/backpressure-high-water-pct) s/Int
+   (s/optional-key :onyx.peer/state-log-impl) StateLogImpl
+   (s/optional-key :onyx.peer/state-filter-impl) StateFilterImpl
+   (s/optional-key :onyx.bookkeeper/starting-port) s/Int
```
MichaelDrogalis: Probably my lack of BK knowledge, but could this just be called "port"? rather than "start-port"?

src/onyx/state/bookkeeper.clj
```diff
@@ -0,0 +1,69 @@
...
+                                 :ledger-dir ledger-dir}))
+                            ports)]
+          (.addShutdownHook (Runtime/getRuntime) 
+                            (Thread. (fn []
+                                       (cleanup-dirs servers))))
+          (assoc component :servers servers)) 
+        component)))
+  (component/stop [{:keys [servers] :as component}]
+    (doseq [server servers]
+      (info "Stopping BookKeeper server")
```
MichaelDrogalis: Needs the equivalent `(info "Starting BK Server")` message on boot up.

src/onyx/state/log/bookkeeper.clj
```diff
@@ -0,0 +1,156 @@
...
+                              (map (juxt :window/id :window/apply-state-update) 
+                                   windows))] 
+    (reduce (fn [state' [window-entries {:keys [window/id] :as window}]]
+              (reduce (fn [state'' [extent entry message-id]]
+                        (update-in state'' 
+                                   [id extent]
+                                   (fn [ext-state] 
+                                     (let [ext-state' (default-state ext-state window)
+                                           apply-fn (id->apply-state-update id)] 
+                                       (assert apply-fn (str "Apply fn does not exist for window-id " id))
```
MichaelDrogalis: Switch from assert to throws.

src/onyx/state/state_extensions.clj
```diff
@@ -0,0 +1,39 @@
+(ns onyx.state.state-extensions)
+
+(defmulti initialise-log
```
MichaelDrogalis: Let's switch the spelling to "initialize" to stay consistent :) <Awkward>

src/onyx/state/state_extensions.clj
```diff
@@ -0,0 +1,39 @@
...
+  "Store state update [segment-id [[extent [op k v]] ..]] entries in a log"
+  (fn [log event ack-fn entry]
+    (type log)))
+
+(defmulti playback-log-entries 
+  "Read state update entries from log"
+  (fn [log event state]
+    (type log)))
+
+(defmulti initialise-filter
```
MichaelDrogalis: Same here.

MichaelDrogalis:
@lbradstreet A truly amazing feat of work here. Added some small comments, but otherwise this looks super. Add any doc updates that are necessary, sync with develop, and merge it in when you're ready. :) Also have a beer!

src/onyx/state/log/bookkeeper.clj
lbradstreet: I made this an assertion because it should always be true and so it can be turned off. I guess performance isn't totally critical here, so maybe we can make it a non-assertion.

src/onyx/state/bookkeeper.clj
lbradstreet: It's already in there on L49

src/onyx/state/state_extensions.clj
lbradstreet: 'murica. Done.

src/onyx/schema.clj
lbradstreet: When you use it in local-quorum? mode, it'll spin up three instances. Generally users will only use one instance per node, so I like port, in general, but maybe we should have an extra config option that specifies which ports to use for a local quorum?

lbradstreet: I added two settings, one which is a vector of ports to use for a local-quorum, and the other to use when not using it in local-quorum mode. At least it's specific now.

src/onyx/state/log/bookkeeper.clj
MichaelDrogalis: Ah, okay. That's fine - wasn't sure if that was the case.

src/onyx/state/bookkeeper.clj
MichaelDrogalis: Derp. Sorry :P

src/onyx/schema.clj
MichaelDrogalis: Cool, that sounds good to me. Thanks for the explanation!

PR #340, opened 2015-10-18
Make interop namespace load other namespaces lazily

### LINKED ISSUE #339: AOT compiled jar causing problems
Hi,

Based on the fact that you are aot compiling `onyx.interop` namespace, I believe the intent of aot compilation is to have some Java classes available in Jar for Java applications?

However, the side effect of AOT is that ALL namespaces that interop namespace transitively depends on are AOT compiled, including clojure.core, timbre, tools.reader... This will cause problems with applications which use need different versions of these dependencies as AOT compiled class files in the classpath will shadow all .clj sources. This completely breaks Clojurescript in project with Onyx dependency.

It should be easy to check what is happening by checking the contents of the jar file.

If the intent of AOT compilation is to provide some Java classes in JAR, I believe the `interop` ns should lazily load other namespaces so that it doesn't have transitive dependencies to other namespaces.

Deraen:
This avoids AOT compilation of complete dependency tree (including clojure.core, tools.reader...) which causes problems with projects which use other versions of these libraries.

Fixes #339

PS. I'm not using Onyx myself so I don't know how interop ns is used. I ran tests but I don't know if they test this.

### DIFF
--- a/src/onyx/interop.clj
+++ b/src/onyx/interop.clj
@@ -2,16 +2,17 @@
   (:gen-class :name onyx.interop
               :methods [^:static [write_batch [clojure.lang.IPersistentMap] clojure.lang.IPersistentMap]
                         ^:static [read_batch [clojure.lang.IPersistentMap] clojure.lang.IPersistentMap]])
-  (:require [onyx.peer.function :refer [write-batch read-batch]]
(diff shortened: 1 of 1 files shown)


MichaelDrogalis:
Fixed on master, release will go out today. Thanks for the patch - ended up going an ever-so-slightly different route.

PR #368, opened 2015-11-04
Use same keyword in description and example

kendru:
On "Flow Conditions", some of the description text referred to `:process-athlete-females`. This changes it to `:process-female-athletes` to make it consistent with the subsequent code snippet.


lbradstreet:
Looks good. Doc fixes are always good :+1

PR #373, opened 2015-11-07
Durable triggers

### LINKED ISSUE #314: Triggers/refinements need to emit log entries
These need to be emitted any time state is refined in a destructive way. 

For example, refine-state :discarding, should write out a "discard" log entry, and then on success do the discard. I'm not sure how we're going to do acking here because these events can happen on a timer.

MichaelDrogalis:
Fixes #314. @lbradstreet ready for review! :)

### DIFF
--- a/src/onyx/peer/task_lifecycle.clj
+++ b/src/onyx/peer/task_lifecycle.clj
@@ -394,14 +394,21 @@
                                             window-id-state (get window-state window-id)
                                             [window-id-state' window-entries] (window-state-updates segment window-id-state window event grouping-fn)
                                             window-state' (assoc window-state window-id window-id-state')]
-                                        (list window-state' 
-                                              (conj log-entries window-entries))))
+                                        (list window-state' (conj log-entries window-entries))))
                                     (list (:state @window-state) [unique-id])
                                     windows)]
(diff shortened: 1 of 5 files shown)


src/onyx/peer/task_lifecycle.clj
```diff
@@ -394,14 +394,21 @@
...
                                     (list (:state @window-state) [unique-id])
                                     windows)]
                         (state-extensions/store-log-entry state-log event ack-fn full-log-entry)
                         (swap! window-state assoc :state new-window-state))
-                      (doseq [t triggers]
-                        (triggers/fire-trigger! event window-state t {:segment segment :context :new-segment})))
+                      (let [trigger-entries
+                            (reduce
+                             (fn [entries t]
+                               (concat
```
lbradstreet: into > concat

src/onyx/peer/task_lifecycle.clj
```diff
@@ -394,14 +394,21 @@
...
+                                entries
+                                [unique-id
```
lbradstreet: We shouldn't store the unique-id a second time, it'll get restored to the filter twice which may cause some issues with bucketing.

MichaelDrogalis:
Fixed. :)

Issue #417, opened 2015-12-03
Add more schema validation to onyx.api functions where appropriate


lbradstreet:
e.g. onyx.api/await-job-completion

MichaelDrogalis:
Fixed in develop.

PR #428, opened 2015-12-09
Add more schema validation to api and cleanup

bridgethillyer:
Is this what you were looking for in #417, @lbradstreet? 

Sorry about the diff-busting indentation cleanup, but @MichaelDrogalis said I could do it. :)


lbradstreet:
Awesome! Apparently I'm the cause of most of the indentation issues. At some point we need to sync up my vim settings with @MichaelDrogalis's emacs settings :/.

PR #432, opened 2015-12-10
Serialize exception that killed a job to ZooKeeper

### LINKED ISSUE #412: Serialize exception that caused kill-job via write-chunk
It'd be useful to look up the exception that caused the job to be killed via the dashboard. This shouldn't cause a large increase in data written to ZooKeeper and it can be GC'd.

MichaelDrogalis:
Fixes #412.

If many tasks fail at the same time, only the first one to write to ZooKeeper will win. All other tasks will fail to write their exceptions to Zookeeper, but will not raise an exception about having failed to do so. This is desired, because ultimately the first task to throw an uncaught exception kills the entire job. Waiting for all tasks to write their exceptions would take an indeterminate amount of time, thus complicating the `onyx.api/await-job-completion-api`. Hence, the job ID is used as the marker for the exception znode.


lbradstreet:
Awesome. Exactly what the doctor ordered. Now for dashboard.

Issue #423, opened 2015-12-07
Ensure peer restarts after ZooKeeper connection loss/errors


lbradstreet:
Found by Jepsen.

Peer does not restart when subscribe-log throws these errors:

* org.apache.zookeeper.KeeperException$ConnectionLossException: KeeperErrorCode = ConnectionLoss for /onyx/JEPSENONYXID/log/entry-0000000105
and
* java.lang.IllegalStateException: Client is not started

Virtual peer should restart on these errors

lbradstreet:
Closed by #484

PR #436, opened 2015-12-10
Ensure peer restarts after ZooKeeper connection loss/errors

lbradstreet:
This solves #423 found by Jepsen. Retest with Jepsen is successful after many runs.

### DIFF
--- a/src/onyx/peer/task_lifecycle.clj
+++ b/src/onyx/peer/task_lifecycle.clj
@@ -520,7 +520,7 @@
           (:onyx.core/triggers event)))
 
 (defn handle-exception [log restart-pred-fn e restart-ch outbox-ch job-id]
-  (warn e)
+  (warn e "task-lifecycle: handle-exception")
   (if (restart-pred-fn e)
     (>!! restart-ch true)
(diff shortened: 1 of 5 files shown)


lbradstreet:
I had forgotton to push to develop earlier so there are a couple commits in here that aren't part of the PR and are already in develop (e.g. midje)

src/onyx/peer/task_lifecycle.clj
```diff
@@ -520,7 +520,7 @@
           (:onyx.core/triggers event)))
 
 (defn handle-exception [log restart-pred-fn e restart-ch outbox-ch job-id]
-  (warn e)
+  (warn e "task-lifecycle: handle-exception")
```
MichaelDrogalis: Is this string that's been added to the end a leftover from debugging?

MichaelDrogalis:
One question about the logging statement. Otherwise looks good!

lbradstreet:
It's kinda nasty to just warn with the exception with no context. I was trying to correct for this. The text could probably be improved though.

PR #443, opened 2015-12-15
Acker daemon tests

MichaelDrogalis:
Beefs up test suite for acking daemon. I chose to omit testing the GC process for now so that I can move on to other tasks, but it's valuable to get this piece in by itself.


lbradstreet:
Looks good

PR #445, opened 2015-12-17
Adds generative test suite for flow conditions

MichaelDrogalis:
This pull request defines a set of composable generators to perform targeted testing of the guts of flow conditions - that which is defined within the task lifecycle itself.

Later added: move flow condition logic into its own file set.

Needs have the rest of the code base fixed up to use the latest test.check and stateful-check to pass the build before merging.


lbradstreet:
Well done :). I simplified some of the generators where we could just `(gen/return {:awhole :map})`. I also extracted and reused the piece of code that merges the two map generators.

MichaelDrogalis:
That's much better, thanks for fixing it up!

Issue #449, opened 2015-12-17
Add configuration for Bookie preallocated space


MichaelDrogalis:
See this discussion: https://groups.google.com/forum/#!topic/onyx-user/J_oSQYu-QMA

MichaelDrogalis:
Patch proposed.

MichaelDrogalis:
Patch merged.

PR #450, opened 2015-12-18
Adds a parameter for disk usage threshold in BookKeeper

MichaelDrogalis:
Adds in the parameter with the corresponding documentation for issue #449.


lbradstreet:
This looks good to me, although there was a reason they added it in the first place - it's probably good practice.

We may want to distinguish between the local-quorum? case, which is clearly testing, and the production case, which is not. Perhaps two settings would be better?

MichaelDrogalis:
I'd vote to keep it at one parameter. Almost all of our arguments exactly match what's provided by the underlying BK API. If the disk starts to fill to capacity, pretty much everything is going to get wonky anyway. Perhaps one other thing we can do is wrap the outgoing exception with a more accurate error message?

lbradstreet:
What I meant is that there was a reason why they made the threshold that low, and now we're defaulting it to a less "safe" threshold (which was chosen to be this high to make dev easier).

Anyway, I think the exception is only thrown when you start up, so it's not like you can't run out of disk space anyway. I guess it doesn't matter so much.

MichaelDrogalis:
Oh - I'm fine with setting the default back to 98%. Even in dev mode, if you're 98% full, you're soon in for trouble. :)

lbradstreet:
This was a typo right? Was the default 90%?

lbradstreet:
I'll add a nicer version of their exception as you suggested.

MichaelDrogalis:
Whoops, yes. Default was 95% to disallow it to start up, 90% to emit a warning. I set it to 98% to disallow start up.

jocrau:
This won't work in cases where somebody tries to set the threshold lower than the warning threshold. This would result in something like:

```
java.lang.IllegalArgumentException: Disk space threashold: 0.8 and warn threshold: 0.9 are not valid. Should be > 0 and < 1 and diskSpaceThreshold >= diskSpaceWarnThreshold
```

I will add a configuration option for the warning and update the PR. (The warn messages are actually the annoying part ;-) )

MichaelDrogalis:
@lbradstreet I will get this merged into develop later today and make a separate issue for wrapping the error with something more pleasant.

Issue #444, opened 2015-12-16
Alter trigger sync function definition


lbradstreet:
New arguments:
event window-map trigger-map {:keys [window-id lower-bound upper-bound]} state

To do by 0.8.4

MichaelDrogalis:
Sailboats issue!

MichaelDrogalis:
Merged this one a while ago.

PR #451, opened 2015-12-18
Update trigger/sync function for new arity.

MichaelDrogalis:
Addresses #444, adds the new trigger arity. Dependent projects will need to be updated after this makes its way into master.


lbradstreet:
Looks good!

PR #457, opened 2015-12-19
Adds a generative test suite for function application

MichaelDrogalis:
Isolates the segment transformation logic and tests features such as bulk application, parameterization, and multi-segment returns.


MichaelDrogalis:
@lbradstreet I'm going to merge this one as I need it to continue other work that depends on it. Let me know if you find anything that we should back out of this patch. It essentially moves some code, unchanged, to another file and adds gen tests.

PR #461, opened 2015-12-25
Add generative tests for flow condition validation

mushketyk:
Added generative tests for the following conditions:

>> If a flow condition specifies **:all** as its **:flow/to**, it must come before any other flow conditions. If a flow condition specifies **:none** as its **:flow/to**, it must come directly behind an :all condition, or first if there is no **:all** condition.


MichaelDrogalis:
I'll check this out when I'm back from vacation on Monday. Thanks!!

MichaelDrogalis:
I think we can change this test a little bit to make it perform many more assertions in a shorter amount of time.

As far as generating test data, on branch `develop`, I recently made a series of [generators for flow conditions](https://github.com/onyx-platform/onyx/blob/develop/test/onyx/flow_conditions/flow_conditions_gen_test.clj#L12-L76). We can factor these into their own file and reuse them for your tests if it helps you.

In terms of performing the assertions, the test as is generates a series of conditions and runs a loop-within-a-loop to scan for correct order - which will yield `n^2` performance. I have a trick that we can use to speed it up.

Instead of generating a random series of flow conditions, it is safe to generate a series for only one task - removing the need to check if the transition matches. Further, we can generate flow condition vectors in the series of `[<matches all>, <matches none>]`. If we let test.check generate two vectors, one with `:flow/to` set to `:all`, and the other set to `:none`, you can concat them together and know exactly which position in the final vector that your flow condition to search for is indexed at, giving you constant time look up.

I wouldn't follow *exactly* the instructions above (haven't had my coffee yet!), but does the idea make sense?

Thanks for generously spending your time on this. :)

mushketyk:
Just to make sure that we understand each other. You suggest, for each test, to generate flow conditions that violate a specific rule and check if all generated flow conditions fail flow conditions validation?

It does make sense, and definitely should work much faster.

MichaelDrogalis:
Exactly :)

mushketyk:
Cool :) I'll work on it in few days.

mushketyk:
Closing to change target branch to "develop".

PR #469, opened 2016-01-04
Add generative tests for flow condition validation

mushketyk:
Added generative tests for the following conditions:

>> If a flow condition specifies **:all** as its **:flow/to**, it must come before any other flow conditions. If a flow condition specifies **:none** as its **:flow/to**, it must come directly behind an :all condition, or first if there is no **:all** condition.


MichaelDrogalis:
Excellent work @mushketyk! Merging. Thanks for contributing :) If you want to continue down this path, you can use the same pattern for other flow condition validations.

mushketyk:
Thank you for merging my PR. I would like to continue down this path. Any advice what is left untested?

MichaelDrogalis:
Position of short-circuiting and exception flow conditions should probably come next. Their positional requirements are described in the same section of the user guide.

mushketyk:
@MichaelDrogalis ok, I'll work on this.

PR #470, opened 2016-01-05
Validations offer suggestions

dignati:
I've added suggestions for workflow keys without matching catalog entries based on existing catalog entries and values of task maps based on the information model.

### DIFF
--- a/src/onyx/static/validation.clj
+++ b/src/onyx/static/validation.clj
@@ -65,24 +65,26 @@
        (take 3)
        (map (fn [suggestion] (keyword (subs suggestion 1))))))
 
+(defn suggest-on-model [model-name key]
+  (let [model (get-in model [model-name :model])
+        choices (not-empty (get-in model [(.value key) :choices]))]
+    (suggest-similar-keys (.value key)
+                          (or choices (keys model)))))
+
 (defn describe-value [k v]
   (if (= schema.utils.ValidationError (type v))
-    (let [doc (dissoc (get-in model [:catalog-entry :model k]) :doc)]
+    (let [doc (dissoc (get-in model [:catalog-entry :model (.value k)]) :doc)]
       (cond
-        (sequential? (:choices doc))
-        {:cause "Unsupported value"
-         :data {k (.value v)}
-         :suggestions (suggest-similar-keys (.value v) (:choices doc))
-         :documentation doc}
-
         doc
         {:cause "Unsupported value"
          :data {k (.value v)}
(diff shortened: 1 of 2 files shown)


dignati:
Not sure where to look next. What's missing?

lbradstreet:
Cool. I haven't tried it, but it looks like it's on the right track.

I think the next most useful one would be to suggest typos in catalog keys that are namespaced onyx. E.g. onyx/m-peers should be onyx/n-peers.

One interesting improvement you could make would be to test whether correcting the mistake with the suggestion passes the schema check, when another doesn't, assuming they both have a close distance (maybe sort by passing validation as a secondary sort tie breaker?).

This could be useful for suggesting something like suggesting onyx/group-by-fn vs onyx/group-by-key since group-by-fn requires a namespaced keyword. Honestly the levenstein distance is probably almost always take precedence.

I would try to make this code general enough that we can use it on all of the following maps: catalogs, lifecycles, aggregations, triggers, etc. 

Thanks, awesome work so far

lbradstreet:
@MichaelDrogalis, elm uses colour judiciously to show exactly where users should glance. This may be a little bit hard for us to do with exception issues. I think as long as we have all the right data in the exceptioninfo, we can do this later or in other tooling if need be. Thoughts?

MichaelDrogalis:
I'd like to see some more focused tests around suggested corrections. This is the type of thing that can have a bug for a long time, but no one would bother to mention it.

lbradstreet:
Yeah we definitely need good tests here. Bad suggestions are probably worse than no suggestions.

dignati:
The last commit should suggest keys in the catalog model.

dignati:
I'm not exactly sure what or how we should test. Testing for specific results could be annoying, those would have to be adjusted every time the model is changed.

dignati:
Sorry, somehow I completely messed up my git repo. I rebased for a clean history.

MichaelDrogalis:
Sorry, I'm a little bogged down. I'll get back to this as soon as I can. Thanks @dignati!

lbradstreet:
Sorry for the lag here on my part too. We're a bit busy but we definitely appreciate the work that you've done!

dignati:
No worries, I'm a bit busy myself. If you tell me what to work on next I
will try to get it done.

Lucas Bradstreet <notifications@github.com> schrieb am Mo., 18. Jan. 2016,
10:36:

MichaelDrogalis:
New error messages are going to approach this in a different way. Thanks for your patch though!

PR #473, opened 2016-01-07
Allow Double values as window key

jocrau:
Using Doubles as window key caused an exception:

```
java.lang.IllegalArgumentException: No implementation of method: :coerce-key of protocol: #'onyx.windowing.units/ICoerceKey found for class: java.lang.Double
```

### DIFF
--- a/src/onyx/windowing/units.clj
+++ b/src/onyx/windowing/units.clj
@@ -106,3 +106,8 @@
   ICoerceKey
   (coerce-key [this units]
     (to-standard-units this units)))
+
+(extend-type java.lang.Double
+  ICoerceKey
+  (coerce-key [this units]
+    (to-standard-units this units)))
--- a/test/onyx/windowing/trigger_watermark_test.clj
+++ b/test/onyx/windowing/trigger_watermark_test.clj
@@ -38,4 +38,22 @@
     (is
      (not
       (api/trigger-fire? event trigger {:window window :segment segment
-                                        :upper-extent (inc t)})))))
+                                        :upper-extent (inc t)}))))
+
+  (let [n 1.0
+        window {:window/id :collect-segments
+                :window/task :identity
+                :window/type :global
+                :window/aggregation :onyx.windowing.aggregation/count
+                :window/window-key :double-number}
+        trigger {:trigger/window-id :collect-segments
+                 :trigger/refinement :accumulating
+                 :trigger/on :watermark
+                 :trigger/sync ::no-op
+                 :trigger/id :trigger-id}
+        segment {:double-number n}
+        event {}]
+    (is
+      (not
+        (api/trigger-fire? event trigger {:window window :segment segment
+                                          :upper-extent (inc n)})))))


MichaelDrogalis:
@jocrau I had to revert this commit. I had a sneaking suspicion that there would be a division error as a result, and I found it with generative testing. Ill revisit this soon, or if you want you can take a look at the commits on develop that show the CI error.

MichaelDrogalis:
@jocrau Can you cast your keys into integers at the SQL import level to work around this for the moment?

jocrau:
Hmm, interesting. Why would there be a division error on Doubles but not on Integers? I will look into it, too.

I tried to coerce it in SQL but that didn't work (the argument can't be an algebraic expression). I will find a work-around for the work-around ;-)

MichaelDrogalis:
@jocrau [Here](https://github.com/onyx-platform/onyx/blob/0.8.x/src/onyx/windowing/window_id.clj#L112-L126) is the relevant code. It can become problematic when the bucketing divides and produces exactly zero buckets for a value, instead of the expected > 0.

jocrau:
@MichaelDrogalis Ok. I see what you mean. I'll take a look on Monday.

jocrau:
@MichaelDrogalis Let' assume we have tuples from a sensor (a totally made up example ;-) ), like

```
{:weight 0.396990501 :color :blue}
{:weight 0.550461931 :color :red}
{:weight 0.692461780 :color :red}
{:weight 1.741405278 :color :blue}
{:weight 2.980380391 :color :green}
{:weight 4.442498492 :color :red}
```

and a window definition

```
WATTR = :weight
RANGE = 2.5
SLIDE = 1.5
```

In this case, the current extent function would yield for w-id = 1:

```
(extents 0 2.5 1.5 1)
=> (0.5 1.5 2.5)
```

but there are no weights 0.5, 1.5, 2.5 in the set of tuples. The result should be:
`(0.550461931 0.692461780 1.741405278 2.980380391)`.

MichaelDrogalis:
@jocrau There are a few things here. Firstly, it isn't semantically correct to window over a value that represents a weight. While the value is monotonically increasing, it doesn't represent a distinct progression over time. There's nothing *wrong* about using the windowing feature with this data, but I don't think it's going to get you the results you were hoping for. `:onyx/group-by-fn` would more likely yield what the sensor situation proposes. Regardless, if we consider the windowing value to be *any* double, we can still talk about what's happening here.

The `extents` function is purposely designed *not* to choose from the set of existing values in a range, because that would require that the set of values be known ahead of time (rarely a realistic assumption), and would require storing all possible values in memory (certainly not possible). I made the same error the first few times I read the paper. The algorithm specifically only calculates the bounds to avoid doing the aforementioned work. By knowing the upper and lower bounds, you can do a `<=` and `>=` comparison to know if a value falls into a particular extent.

jocrau:
@MichaelDrogalis The term "weight" might not be the best choice, but the only thing that matters is, that windowing should work with any (strict or non-strict) monotonic sequence of values (out-of-order arrival is a different challenge).

Let's step back and just assume we have a monotonic sequence of integer values (like milliseconds, and to make things easier we start from 1). We might not have a tuple every millisecond, so our stream of tuples might look like

```
{:event-time 1 :some-attr 12}
{:event-time 3 :some-attr 10}
{:event-time 7 :some-attr 7}
{:event-time 8 :some-attr 18}
{:event-time 10 :some-attr 9}
{:event-time 11 :some-attr 11}
```

With

```
WATTR = :event-time
RANGE = 4
SLIDE = 2
```

we currently get

```
(extents 0 4 3 2)
=> (5 6 7 8)
```

which is clearly wrong. The correct sequence of values would be `(7 8)`. But even is the values were correct, returning the sequence of values is not sufficient. Since we cannot rely on :event-time being a unique, bijective identifier (although we can ensure that by specifying :onyx/uniqueness-key in other cases), we cannot "query" the tuple using it. To be useful, the `extent` function should return a sub-set of tuples (`t ∈ T`) that fall within a window. Thus, the correct result should be

```
[{:event-time 7 :some-attr 7}
 {:event-time 8 :some-attr 18}]
```

which would require to hand in the set of tuples as an argument to the function as the paper suggests. I am not sure, but isn't the set of tuples `T` to iterate over the same as the collection this code reduces https://github.com/onyx-platform/onyx/blob/0.8.x/src/onyx/peer/task_lifecycle.clj#L249 ?

I agree that the upper and lower bounds of the extent is useful for determining whether a specific tuple is within an extent of a given window. AFAICS the `extents` function is never used (other than as test data generator). The `onyx.windowing.window-extensions/extents` that is used by `onyx.peer.task-lifecycle/window-state-updates` seems to return the ids of the windows a specific tuple falls into (it finally calls `onyx.windowing.window-id/wids`).

P.S.
The `:onyx/group-by-fn` won't work (in the constructed example), as I don't want to group the tuples by weight.

MichaelDrogalis:
@jocrau I'm all for making changes to the algorithm if my implementation of the paper isn't faithful to the specification, but this misses the mark. I think you need to read the paper again if you want to propose a change here. Furthermore, the suggested change about `extents` needing the tuple domain is independent of the type of window key - there is nothing here that gets closer to allowing decimal types as the window key. I don't think it's wrong to allow double types - but I know it needs more careful thought in the area of code that is doing division.

> out-of-order arrival is a different challenge

It's not - the point of the WID paper is to be able to efficiently bucket data in an order-of-order fashion. This is stated as one of the main goals of the research in the first paragraph of the paper.

> The correct sequence of values would be (7 8).

This is not so. The paper is explicit in what needs to happen when extents "slide in from the right of the minimum". Windows are of *partial* length until they can fit in the entire domain of the window. This can also be cross-checked against their example.

> Since we cannot rely on :event-time being a unique, bijective identifier (although we can ensure that by specifying :onyx/uniqueness-key in other cases), we cannot "query" the tuple using it. To be useful, the extent function should return a sub-set of tuples (t ∈ T) that fall within a window. 

You're asking `extents` to do something it wasn't meant to do. I don't know what "useful" is for your use case, but it clearly serves the purpose of bucketing - further described as one of the cornerstone operations in the Google DataFlow paper.

> FAICS the extents function is never used (other than as test data generator).

Indeed, it's used to check the implementation of `wids`. It is later used in the paper as the basis for more algorithms that we may pull in at a later time.

jocrau:
I know where the confusion might come from: The PR and commit was titled "Allow Double values as window key", but in fact the change that is needed is to allow Doubles (or any kind of comparable value) as `t.WATTR`. That is what the proposed code change actually addresses. I changed the title now. Sorry about that.

I read the paper quite carefully, and I read the definition of the `extent` function in section 3.3 again as you suggested ;-).

I am aware that handling out-of-order tuples is one of the topics of the paper. But I am focusing on the semantics of `extent`, which the paper defines as (assuming `min WATTR(T ) = 0` to make things easier):

```
extent (w,T,S[RANGE, SLIDE, WATTR]) = { t ∈ T | (w+1) ∗ SLIDE − RANGE ≦ t.WATTR < (w+1) * SLIDE }
```

Here is how I interpret the definition: given a window id `w`a set of tuples `T` and a window specification `S`, determine the set of tuples `{ t ∈ T }` for which the value of the window attribute (`t.WATTR`) is between (inclusive) the lower boundary of the extent (`(w+1) ∗ SLIDE − RANGE`) and the upper (exclusive) boundary of the extent (`(w+1) * SLIDE`).

The definition makes no assumption about the order of tuples in `T`. That's why I suggested out-of-order arrival is a different challenge (addressed in section 7).

There might be an argument to make whether the `extent` function should return the set of tuples (`t ∈ T`) or the set of values (`{v | v=x.WATTR, ∀x ∈ T}`). I personally would go with the paper (how I interpret it) and return the tuple, because we cannot assume that the value is a bijective identifier.

The question whether the interpretation of `extent` is correct or not goes a bit beyond the purpose of this PR, and because it is currently not used, it will not create any problems in the short term. But I would like to get the window semantics correct in my head and avoid future issues. That's why I am a bit persistent here ;-) 

I am just beginning to look into the internals of Onyx and certainly not familiar with all the details. So, could you please explain to me what is the (future) purpose of the `extent` function (plural or not) is and what it is supposed to return in this context?

I am also open to have a quick chat on Skype if you want.

P.S. Division by Zero wasn't the issue. Coercing a Double into a long didn't work without setting a max in the generator (fixed with this [commit](https://github.com/jocrau/onyx/commit/7088ee0cfcee8b90dcf0b0fdd757b6bfc4a86aa0)).

PR #479, opened 2016-01-11
Use SHARED Aeron media driver by default

mushketyk:
I see that unit tests execute changed parts of code, but how do you think, should I run some additional tests?

### DIFF
--- a/src/onyx/information_model.cljc
+++ b/src/onyx/information_model.cljc
@@ -824,6 +824,13 @@
              :default 1000
              :added "0.8.0"}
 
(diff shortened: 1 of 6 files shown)


lbradstreet:
I'm going to merge this but I'm going to change a few things afterwards, e.g. the peer-config param name and make it use a keyword.

lbradstreet:
Thanks. I made some improvements here, if you're curious https://github.com/onyx-platform/onyx/commit/c64683694612a78d250881341a5866b81dc42039

PR #486, opened 2016-01-16
Perform java version validation above onyx.api ns

lbradstreet:
This allows us to perform a check without resorting to dynamic loading
of aeron messaging, which actually causes a number of painful delayed
error messages.


MichaelDrogalis:
A user can still get burned by requiring a namespace other than `onyx.api` first. It's unlikely, but it can happen. I agree with you on the other thread - the best thing to do is let it stand.

PR #518, opened 2016-02-01
Fix an API documentation link

reaperhulk:
Looks like one link was missed.


MichaelDrogalis:
@reaperhulk Thanks!! :)

PR #523, opened 2016-02-06
Constraint fixes

MichaelDrogalis:
Fixes issues that we found with constraints, @lbradstreet. This now uses the Fence constraint to make a peer either pick its existing task, or any tasks with its matching tag.


lbradstreet:
Hi @MichaelDrogalis, I improved the generative tests (one of which fails). I also converted the (remove nil? (map XX XX)) cases to use "keep" to reduce the nesting.

lbradstreet:
I added another similar failing test for grouping, which makes me think that both the fencing and quarantining break reallocations on kill-job.

See onyx.scheduler.balanced-grouping-generative-test/recover-slots

lbradstreet:
I added a bunch of replica invariants to help with this process. The assertions now fail quite early, though we get a lot more failing tests now. Most appear to do with grouping and constraints tags, so I believe the  issue is the same in all of them.

MichaelDrogalis:
Fixed. I'm going to merge this immediately since we discovered an important fix. I'm happy to keep banging on this feature with more tests though.

PR #526, opened 2016-02-11
Boot out dead peers when the replica doesn't reflect the cluster state.

MichaelDrogalis:
@lbradstreet Up for review. We should reproduce this one before adding a patch, as you said.

### DIFF
--- a/src/onyx/log/commands/prepare_join_cluster.clj
+++ b/src/onyx/log/commands/prepare_join_cluster.clj
@@ -34,7 +34,7 @@
   (let [all-prepared-deps (keys prepared)
         prep-watches (map (fn [dep] (get (map-invert pairs) dep)) all-prepared-deps)
         accepting-deps (keys accepted)]
-    (set (concat all-prepared-deps accepting-deps prep-watches))))
+    (concat all-prepared-deps accepting-deps prep-watches)))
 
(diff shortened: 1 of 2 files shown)


src/onyx/log/commands/prepare_join_cluster.clj
```diff
@@ -93,10 +93,25 @@
...
 (s/defmethod extensions/fire-side-effects! :prepare-join-cluster :- State
-  [{:keys [args message-id]} :- LogEntry old new diff {:keys [monitoring] :as state}]
+  [{:keys [args message-id]} :- LogEntry old new diff {:keys [log monitoring] :as state}]
   (common/start-new-lifecycle
    old new diff
-   (cond (= (:id state) (:observer diff))
+   (cond ;; Handles the cases where all peers are actually dead.
+         ;; This can happen if a single node cluster comes down
+         ;; and is rebooted. We pick a predictably-random peer
+         ;; and knock it down if it's not up. This garuntees
```
thenonameguy: s/garuntees/guarantees

lbradstreet: Thank you :D

lbradstreet:
The cond was checking whether the peer was in the diff, which was nil. I switched it to check the log entry. I also made it evict accepted and pairs.

Jepsen approved, and thus merging! :+1:

PR #531, opened 2016-02-18
No need to assoc workflow to job.

tiensonqin:
Since every job should have the key `:workflow`.


MichaelDrogalis:
Hi, thanks for this patch. Can you open it against the master branch, please?

tiensonqin:
Sure!

PR #532, opened 2016-02-19
No need to assoc workflow to job.

tiensonqin:
Since every job should have the key `:workflow`.


MichaelDrogalis:
Thank you @tiensonqin! :)

Issue #529, opened 2016-02-13
Make refinement and aggregations state fn names and signatures consistent


lbradstreet:
e.g. apply-state-update / create-state-update

Issue #530, opened 2016-02-13
Fix timed trigger checkpointing


lbradstreet:
Must happen at a safepoint in the task-lifecycle

Issue #520, opened 2016-02-04
Merge extents should generate log entries which are written to changelog


lbradstreet:
Use of merge-extents is unsafe because it mutates the windows without a corresponding log entry.

When the peer reboots and replays the changelog, there are no corresponding window updates that merge the session windows.

lbradstreet:
Also I believe speculate-updates is also guilty of this.

Issue #519, opened 2016-02-04
Timer trigger is unsafe when used with mutating refinements


lbradstreet:
Storing log entries from trigger times is unsafe when used with any kind of mutating refinement as it does not run at a safepoint in the task-lifecycle.

lbradstreet:
This will be fixed in 0.9.0 triggers

PR #533, opened 2016-02-21
Improve state management and triggers #529 #530 #519 #520 #510 #507

lbradstreet:
Still required:
Trigger state scoping e.g. there's currently a trigger state maintained for each window e.g. every 5 segments for a window will cause the window to fire, but it may be the case that you want to fire every 5 segments for all the windows, or possibly every 5 segments for any grouping key.

Closes: 
#529 #530 #519 #520 #510 #507

Still left to be done:
Parameterizable notifications #488


lbradstreet:
I just realised that https://github.com/onyx-platform/onyx/blob/0.8.x/doc/user-guide/triggers.md is out of date now. Fix this up before we merge.

src/onyx/information_model.cljc
```diff
@@ -380,18 +380,45 @@
...
                                :added "0.8.0"}
-            :aggregation/fn {:doc "Fn (state, window, segment) to generate a serializable state machine update."
-                             :type :function
-                             :optional? false
-                             :added "0.8.0"}
-            :aggregation/apply-state-update {:doc "Fn (state, entry) to apply state machine update entry to a state."
+            :aggregation/create-state-update {:doc "Fn (window, state, segment) to generate a serializable state machine update."
+                                              :type :function
+                                              :optional? false
+                                              :added "0.8.0"}
```
MichaelDrogalis: Let's bump the version to 0.9.0.

src/onyx/information_model.cljc
```diff
@@ -380,18 +380,45 @@
...
                                                :optional? true
                                                :added "0.8.0"}}}
+
+   :state-refinement
+   {:summary "Onyx provides the ability to perform state refinements after triggers fired."
+    :link nil
+    :model {:refinement/create-state-update {:doc "Fn (trigger, state, state-event) to generate a serializable state machine update."
+                                              :type :function
+                                              :optional? false
+                                              :added "0.0.0"}
```
MichaelDrogalis: Switch version.

src/onyx/information_model.cljc
```diff
@@ -380,18 +380,45 @@
...
+            :refinement/apply-state-update {:doc "Fn (trigger, state, entry) to apply the refinement state machine update entry to a state."
+                                             :type :function
+                                             :optional? false
+                                             :added "0.8.0"}}}
```
MichaelDrogalis: Version.

src/onyx/information_model.cljc
```diff
@@ -658,6 +696,13 @@ You can use lifecycles to supply any stateful connections necessary to sync your
              :default []
              :added "0.8.9"}
 
+            :onyx.peer/trigger-timer-resolution
+            {:doc "The resolution of the timer firing state-events that are not caused by segments arriving."
+             :type :integer
+             :optional? true
+             :default 100
```
MichaelDrogalis: Let's add a `:units` key here. Milliseconds, I presume?

src/onyx/peer/task_lifecycle.clj
```diff
@@ -368,6 +323,16 @@
             (extensions/write-chunk log :exception e job-id)
             (>!! outbox-ch entry))))))
 
+(s/defn assign-windows :- Event
```
MichaelDrogalis: This area got so much cleaner. Nice work.

src/onyx/information_model.cljc
lbradstreet: These were present in 0.8.0. You're right about the others though, I'm not sure how I missed them.

src/onyx/triggers.clj
```diff
@@ -0,0 +1,144 @@
+(ns onyx.triggers
```
MichaelDrogalis: Can we move this file back down into the triggers directory?

MichaelDrogalis:
Looks good on first pass, but definitely need to talk about it during a review when we're both around. Clear to merge. :+1:

PR #536, opened 2016-02-27
Job metadata

MichaelDrogalis:
Adds metadata to job and improves logging.

### DIFF
--- a/doc/user-guide/apis.md
+++ b/doc/user-guide/apis.md
@@ -26,7 +26,7 @@ Starts N virtual peers to execute tasks. In a production environment, you should
 
 ##### `submit-job`
 
-Submits a job to Onyx to be scheduled for execution. Takes a map with keys `:catalog`, `:workflow`, and `:task-scheduler`. Returns a map of `:job-id` and `:task-ids`, which map to a UUID and vector of maps respectively.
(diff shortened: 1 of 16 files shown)


lbradstreet:
I made one small change to remove unused lookups.

The main problem with the new code is that it doesn't allow us to elide the calls completely, so the code in the trace here, e.g. the count, will be evaluated even though task-log-trace will have an empty function body with TIMBRE_LOG_LEVEL=trace.

```
(logger/task-log-trace
     task-information
     (format "Wrote %s segments" (count (:onyx.core/results rets))))
```
I think what I'd rather do is pre-build the log prefix formatted by the fn below and stick that string in a key in the compiled record. Then we look that string up inside the info/debug/etc macro calls, so even the lookup will be elided with the right TIMBRE_LOG_LEVEL.

```
(defn log-prefix [task-info]
  (format "Job %s %s - Task %s %s - Peer %s - "
          (:job-id task-info)
          (:metadata task-info)
          (:task-id task-info)
          (:task-name task-info)
          (:id task-info)))
```

lbradstreet:
I fixed that issue and rebased on develop, requiring a force push. You'll get an error if you pull but I think it's good to go.

lbradstreet:
Merging for 0.9.0 alpha release.

PR #537, opened 2016-02-28
Remove duplicated default values.

MichaelDrogalis:
Derives the default values from the information model. I checked and ensure that all the values in the defaults file match what's in the information model. cc @lbradstreet


lbradstreet:
Looks good to me! Thanks. I've been meaning to do that for a while.

PR #541, opened 2016-03-10
reorganized core_async util functions

gardnervickers:


gardnervickers:
This is going to take a bit of coordination. Right now all the plugins are depending on `0.9.0-alpha9`. I will move `onyx.plugin.core-async-tasks` to `onyx.tasks.core-async` to fit with how redis/kafka/datomic are now, but we need another alpha release after I make that change.

PR #543, opened 2016-03-16
Schema changes to drop deprecated ->Both and

gardnervickers:
improve readability of error messages by applying value checks as a postcondition. I tried chunk removing from the commit all line ending/formatting related code.


lbradstreet:
Looks good. Merge into develop in the future so we don't have to merge from master to develop though. I'll manually merge via git.

Issue #539, opened 2016-03-08
handle-exception does not apply to exceptions thrown in flow conditions


lbradstreet:
This has come up in the wild. Sometimes you have unrecoverable exceptions in your flow condition predicates, etc and wish for the peers to restart rather than have your job reboot. This issue is not *totally* clear cut, so should discuss whether it's a good idea.

MichaelDrogalis:
Code, test, and docs ready on this one. @lbradstreet

MichaelDrogalis:
Merged.

PR #547, opened 2016-03-16
Solves #539, allowing lifecycles to catch flow condition exceptions.

MichaelDrogalis:
Flow condition exceptions can now be caught via lifecycle exception handlers.


lbradstreet:
Fine to merge after incorporating the test into lifecycles-exception-test

MichaelDrogalis:
@lbradstreet I looked at merging these two tests, and it's relatively complicated. The lifecycles-exception-test throws lifecycles at every stage of the lifecycle and forcibly continues all the way to job completion. I tried to add flow condition exceptions. The trouble is that when you do a throw inside of a flow condition, you lose the segment if it unravels all the way lifecycle exception handler. You can try to make a guess about which segments shouldn't show in the core.async channel on the other end, but I'm not sure that we'd be able to do that reliably. I'm going to merge as is, but I'll sleep on a way to merge these together as I agree that it inflates the test suite size, which hurts.

Issue #534, opened 2016-02-22
Should handle-exception be called for apply-fn?


lbradstreet:
There are some exceptions thrown by onyx/fn that should not be handled by flow conditions. :lifecycle/handle-exception should probably be able to handle these (e.g. an exception thrown by a bad connection injected into the event map).

The main issue is that it's a fairly blunt force tool, that will lead to segments not caused by transient exceptions being retried over and over.

lbradstreet:
Fixed in #548

PR #548, opened 2016-03-17
Apply fn exceptions

MichaelDrogalis:
Solves #534, lifecycles can catch apply-fn exceptions. See Slack for explanation of the extra test. @lbradstreet


lbradstreet:
This is good now that I've fixed the failing flow cond gen test.

PR #549, opened 2016-03-17
Merge schema between user-defined schema and core

gardnervickers:
This PR is to make it so that users of task bundles dont have to supply the `base-schemas`.


lbradstreet:
Looks good, thanks.

PR #553, opened 2016-03-19
Implement more fine grained trigger signals e.g. :job-completed

lbradstreet:
This implements better trigger event types and schema checks for them. Triggers and sync fns can now tell whether the event results from job completion, peer scheduling, etc.

No changes to information_model.cljc are required, but we should probably have a way to link stuff like types taken by functions to our schemas.


lbradstreet:
Needs to be documented in CHANGES.MD

MichaelDrogalis:
Excellent, excellent, excellent. :+1:

lbradstreet:
Merged. Going to add trigger events to the information model and revamp the way schema works in another branch.

PR #570, opened 2016-04-13
Fixes #554 by constraining the amount of output on select records.

### LINKED ISSUE #554: Create print-method for peer component
Error messages end up pages long after they print the whole peer component. I think that we should be printing less of the component for consumption by users.

e.g. something like this (stolen from component)

```clojure
#?(:clj
   (defmethod clojure.core/print-method SystemMap
     [system ^java.io.Writer writer]
     (.write writer "#<SystemMap>"))
   :cljs
   (extend-protocol IPrintWithWriter
     SystemMap
     (-pr-writer [this writer opts]
       (-write writer "#<SystemMap>"))))
```

MichaelDrogalis:
Alters some records to have quieter outputs.

### DIFF
--- a/src/onyx/log/zookeeper.clj
+++ b/src/onyx/log/zookeeper.clj
@@ -132,6 +132,10 @@
 
     component))
 
+(defmethod clojure.core/print-method ZooKeeper
+  [system ^java.io.Writer writer]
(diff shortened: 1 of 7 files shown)


lbradstreet:
Looks good.

PR #576, opened 2016-04-21
Show friendlier error message for schema validation

bridgethillyer:
* we are already doing so for catalog, add to peer-client-config,
flow conditions, lifecycles, windows, and triggers


MichaelDrogalis:
Closing this in favor of the new static analysis. Thanks though!

PR #580, opened 2016-04-24
6 keys/functions/etc seems to be 9 now

jackdempsey:
Was a bit confusing, seems just like a typo.

### DIFF
--- a/doc/user-guide/lifecycles.md
+++ b/doc/user-guide/lifecycles.md
@@ -52,7 +52,7 @@ If an exception is thrown during any lifecycle execution except `after-task-stop
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


jackdempsey:
Hmm, I'm realizing only 6 functions actually do return a map, so maybe the fix is to change other wording? Anyway, some clarification here would be good.

MichaelDrogalis:
@jackdempsey I'll reword this and post back on completion. Thanks for the report!

MichaelDrogalis:
@jackdempsey Sorry this took a while to get to. I just went ahead and fixed it in 35049f666f9be57727cee2dd8b8cd0fdb526b888 since the patch needs to go to develop first. Thanks for bringing it up!

jackdempsey:
Awesome @MichaelDrogalis thanks!

PR #582, opened 2016-04-29
Static Job Analysis

MichaelDrogalis:
Provides significantly more readable errors for static job analysis.


MichaelDrogalis:
@gardnervickers @lbradstreet Ready for review.

src/onyx/api.clj
```diff
@@ -145,40 +145,52 @@
         args (add-percentages-to-log-entry config job args tasks (:catalog job) id)]
     (create-log-entry :submit-job args)))
 
+(defn validate-submission [job peer-client-config]
+  (try
+    (validator/validate-peer-client-config peer-client-config)
+    (validator/validate-job job)
+    (validator/validate-flow-conditions job)
```
lbradstreet: I believe we can drop all these additional cases if validate-job did its job correctly. I've fixed this in a commit I'll push up.

src/onyx/static/analyzer.clj
```diff
@@ -0,0 +1,173 @@
...
+
+               (= (type form) schema.utils.NamedError)
+               (assoc result path (classify-error job path (.-error form)))
+
+               (= form 'missing-required-key)
+               (assoc result path {:error-type :missing-required-key
+                                   :path path
+                                   :missing-key (last path)})
+
+               (= form 'disallowed-key)
```
lbradstreet: Added this case, even though it doesn't currently show up in our schemas

src/onyx/static/helpful_job_errors.clj
```diff
@@ -0,0 +1,858 @@
+(ns onyx.static.helpful-job-errors
+  (:require [clojure.string :refer [split join]]
+            [onyx.information-model :refer [model]]
+            [clj-fuzzy.metrics :refer [levenshtein]]
+            [io.aviso.ansi :as a]))
+
+(def manual-ex (ex-info "Manual validation check failed" {}))
+
+(def structure-names
```
lbradstreet: should we just rename these in the information model? The mapping is a bit awkward and we might as well just name them what they should be. That said, catalog-entry / task-map both make more sense than :catalog because the catalog is the full list of task-maps

lbradstreet:
Looks good to me.

One thing we may want to consider in the future is not throwing when we call print-helpful-job-error-and-throw, and instead build up all of the errors. This would allow us to show all of the errors in a job at once, which would allow a user to fix more errors at once, and allow them to diagnose things better when they have two interacting issues, rather than bailing on the first issue.

src/onyx/static/helpful_job_errors.clj
MichaelDrogalis: I'm okay with keeping it. I'd rather not disrupt the information model file.

MichaelDrogalis:
Agreed, it'd be nice to have a print out of "You have N errors. Here are each of them."

PR #584, opened 2016-05-11
Multiplex zk conn

MichaelDrogalis:
Uses one ZooKeeper connection across all peers. @lbradstreet ready for review, needs to pass Jepsen tests.


lbradstreet:
I *think* this looks good. My only concern is that the pulse might not be unregistered on leave cluster, but since the peer will shutdown the processing loop, I *think* it's good.

Jepsen will say for sure since it checks the pulses match the replica afterwards.

MichaelDrogalis:
Possible failure case -

Peers connect successfully to ZK, then they partition. Peers cant deregister their pulses, so that fails. The group restarts its peers, but just before that the partition heals. Peers see old pulses that will never go away because the connection wasn't explicitly closed.

We can scoot around this by maintains the previous peer ID and deregistering on boot up if it exists.

MichaelDrogalis:
@lbradstreet This is ready for code review + Jepsen testing. Will squash merge the work into develop when approved.

lbradstreet:
Short note for later explanation. We really need to switch all the other generative tests to use groups that join independently. Otherwise we're just going to get a single group join, then N peer joins without any interesting behaviour from the scheduler.

MichaelDrogalis:
I can do this today.

lbradstreet:
Merged manually

PR #588, opened 2016-05-27
Don't preallocate peers to grouped tasks with flux-policy continue

dignati:


dignati:
Tests fail because of job-upper-bound in the common -job-scheduler

lbradstreet:
Looks good to me. I'll merge and work on the tests from within develop. Thanks! 🎉

PR #591, opened 2016-05-29
Allow for multi-key :onyx/uniqueness-key if expressed as a vector

### LINKED ISSUE #590: Allow multi-key uniqueness-key
I envision something like passing a vector of keywords instead of a single keyword:

```
:onyx/uniqueness-key [:page :position]
```

And internally it would perhaps `juxt` the getter results (as you would do in sorting on multiple keys). This would save having to create an artificial key (i.e. [something that can be used as an argument to `get`](https://github.com/onyx-platform/onyx/blob/686ca2a8bb1fed8cd3a6dbe63d75923460f46888/src/onyx/peer/window_state.clj#L240)) just for windowing needs. However, I'm not sure if this idea creates problems for how this is tracked, but [it looks like it would work](https://github.com/onyx-platform/onyx/blob/cf577d415a9b3f4016cfbf8675fb751a1671bb8d/src/onyx/state/filter/rocksdb.clj#L85)?

neverfox:
Closes #590

### DIFF
--- a/doc/user-guide/aggregation-state-management.md
+++ b/doc/user-guide/aggregation-state-management.md
@@ -122,12 +122,12 @@ while remaining correct on peer failure.
 ### Exactly Once Aggregation Updates
 
 Exactly once aggregation updates are supported via Onyx's filtering feature. When a
-task's catalog has `:onyx/uniqueness-key` set, this key is looked up in the
-segment and used as an ID key to determine whether the segment has been seen
-before. If it has previously been processed, and state updates have been
-persisted, then the segment is not re-processed. This key is persisted to the
-state log transactionally with the window changelog updates, so that previously seen keys
-can be recovered in case of a peer failure.
+task's catalog has `:onyx/uniqueness-key` set, this key (or vector of keys, in the case
+of multi-key uniqueness) is looked up in the segment and used as an ID key to determine
+whether the segment has been seen before. If it has previously been processed, and state
+updates have been persisted, then the segment is not re-processed. This key is persisted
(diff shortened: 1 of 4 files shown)


lbradstreet:
A few things that I can see here. You don't necessarily need to fix them @neverfox, but they will require a fix before we land it. 

1. We need to precompile to a uniqueness fn so that we're not doing the apply flatten juxt for every segment. This is both for performance reasons and to keep the task logic in one place.
2. Get rid of juxt and use "get segment key", because the keys may not be keywords.
3. Get rid of flatten, instead check whether it's a sequential or not when we precompile the fn (see what we do with group-by-key/fn here.

Once we have the uniqueness-key work precompiled to a function used everywhere it becomes very easy to add "onyx/uniqueness-fn" too.

neverfox:
@lbradstreet All very good points to improve on this quick-and-dirty solution. I particularly like the idea of an `:onyx/uniqueness-fn`.

lbradstreet:
Too many things are changing in the next version, and the main need for this feature was somewhat of an abuse of the feature, which was simply intended to deduplicate segments when they are retried. This will not be so necessary in the next version. If you need something like this, could you please describe your use case in an issue so that we can cover it in the next version? The next version of Onyx won't need to filter / deduplicate retries, so it won't be as necessary to have a deduplication filter, but I am open to use cases which aren't due to message retries.

neverfox:
I can wait for the next version. We're currently using v5 UUIDs to create uids for retries.

PR #592, opened 2016-06-05
Multiplex peer group connections via single writer

lbradstreet:
This multiplexes peer groups via single writer. Includes a generative test to put the group through a series of changes and check for basic properties.

Also changes the reactions / fire-side-effects! multimethods to take a vector of [message-type actor-type], where actor-type is :peer/:group/:client. Client being for calls like onyx.api/gc.

@MichaelDrogalis please review these, and we will then do one big final review on the main PR request.

Note that many tests now require await-job-completion/feedback-exception!, as the peers shutdown too fast to finish the job once the :done is read. This is acceptable, as there are no guarantees otherwise anyway.


lbradstreet:
Merged manually

PR #597, opened 2016-06-15
Fix it's -> its on `kill-job`

devth:


MichaelDrogalis:
Thanks @devth. Build failed because S3 logs failed to upload on CI. Patch is safe.

PR #598, opened 2016-06-15
Idempotent job submission, fixes #573

### LINKED ISSUE #573: Make job submission idempotent
Ideally the way this would work is you would supply the job-id to submit job. If the job has already been submitted (with all chunks written), then a second submit-job would have no effect. This may or may not require extra log entries that are treated idempotently (probably not). If there are extra log entries you have to be careful that the job doesn't get re-run if the log is GC'd.

MichaelDrogalis:
Documentation is still needed. Solves #573.

### DIFF
--- a/src/onyx/api.clj
+++ b/src/onyx/api.clj
@@ -208,11 +208,11 @@
    (let [result (validate-submission job peer-client-config)]
      (if (:success? result)
        (let [job (update-in job [:metadata :job-id] #(or % (UUID/randomUUID)))
-             job-hash (hash-job job)
              id (get-in job [:metadata :job-id])
              tasks (planning/discover-tasks id (:catalog job) (:workflow job))
              entry (create-submit-job-entry id peer-client-config job tasks)
              client (component/start (system/onyx-client peer-client-config monitoring-config))
+             job-hash (hash-job job)
              status (extensions/write-chunk (:log client) :job-hash job-hash id)]
          (if status
            (serialize-job-to-zookeeper client id job tasks entry)
--- a/src/onyx/api.clj
+++ b/src/onyx/api.clj
@@ -7,11 +7,8 @@
             [onyx.extensions :as extensions]
             [onyx.static.validation :as validator]
             [onyx.static.planning :as planning]
-            [onyx.static.default-vals :refer [arg-or-default]]
-            ;; leave-cluster must be imported through api.clj,
-            ;; not system.clj like all the other log entries to
-            ;; prevent a cyclic namespace dependency.
-            [onyx.log.commands.leave-cluster]))
+            [onyx.static.default-vals :refer [arg-or-default]])
+  (:import [java.util UUID]))
 
 (defn ^{:no-doc true} saturation [catalog]
   (let [rets
@@ -183,7 +180,8 @@
   ([peer-client-config job monitoring-config]
    (let [result (validate-submission job peer-client-config)]
      (if (:success? result)
-       (let [id (or (get-in job [:metadata :job-id]) (java.util.UUID/randomUUID))
+       (let [job (update-in job [:metadata :job-id] #(or % (UUID/randomUUID)))
+             id (get-in job [:metadata :job-id])
              tasks (planning/discover-tasks (:catalog job) (:workflow job))
              client (component/start (system/onyx-client peer-client-config monitoring-config))]
          (extensions/write-chunk (:log client) :catalog (:catalog job) id)
--- a/src/onyx/api.clj
+++ b/src/onyx/api.clj
@@ -174,21 +174,21 @@
     (let [digest (.digest md)]
       (apply str (map #(format "%x" %) digest)))))
 
-(defn ^{:no-doc true} serialize-job-to-zookeeper [client id job tasks entry]
-  (extensions/write-chunk (:log client) :catalog (:catalog job) id)
-  (extensions/write-chunk (:log client) :workflow (:workflow job) id)
-  (extensions/write-chunk (:log client) :flow-conditions (:flow-conditions job) id)
-  (extensions/write-chunk (:log client) :lifecycles (:lifecycles job) id)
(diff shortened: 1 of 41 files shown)


lbradstreet:
I haven't had a chance to run this code yet, but it looks nice overall. My main concern is that the second run will have different task ids, to the first run, and the original, correct, task ids with possibly allocated peers will be now out of date.

If this is handled correctly (maybe it simply doesn't force write the chunk with that data), then all is well.

MichaelDrogalis:
Your last statement is true. We always try to write a chunk to the `tasks/` directory. Only the first one will succeed, all others will fail and read the original task ids, returning them instead. I'll add tests and docs today.

MichaelDrogalis:
Found a bug. Hold on review.

MichaelDrogalis:
Updated with fix, plus tests and docs. Ready for review.

lbradstreet:
If we're going to allow users to stick the job-Id in the metadata, I think maybe we should stick a generated UUID into the metadata when one is not supplied. This will make it easier for other code to use job ids from the metadata.

MichaelDrogalis:
I think the job ID is particularly more important than the rest of the metadata, so I'd be inclined to keep it stored as a first class value. Moving the job ID into another location would require quite a few code changes. We can discuss though.

lbradstreet:
What I'm saying is that we should stick the job id into the metadata for users before we write out the metadata to zookeeper - it should be a single line change.

We need to also validate the supplied job data that  {:metadata {:job-id xx}} is either a UUID or doesn't exist. Otherwise users could supply non UUID job ids and it'll only fail when peers startup, and only if they're using schema validation (it'll see that the job id is incorrect in the event, but this could possibly just cause the peer to reboot given handle-exception lifecycles).

lbradstreet:
My suggestion above will also allow us to ditch the or in task lifecycle:


          metadata (or (extensions/read-chunk log :job-metadata job-id) {})

MichaelDrogalis:
Apologies, I thought you were suggesting something much different. Good to me.

I intentionally widened the schema to allow for non-UUID job-IDs, since I think it would be useful to pass in plain strings. I should narrow it back down somewhat, though. Thoughts?

lbradstreet:
submit-job diff shouldn't emit a diff if nothing changed. This is won't change how anything works - it's more for nice reporting.

Should we be worried about gc'ing the replica? My feeling is that we should stop clearing killed and completed jobs - I'd rather have it be safe vs ensuring we reduce the size of the replicas as much as possible

lbradstreet:
I think we should leave the job-id as a UUID and use job metadata to allow for the sort of querying that users might want. Otherwise we'll end up with users using names like fooble-stream1, fooble-stream2, etc, as they update their job. I'd rather something that doesn't collide and then use the metadata for that stuff.

lbradstreet:
Another problem is something like this. You have two machines generate a job each, using the same job id. Now, they each generate different jobs because they built their job using something global and mutable e.g. a DB. Some of the chunk writes succeed from one submit, some from the other. Now you have an inconsistent job definition, even though each of the individual jobs were consistent and fine.

lbradstreet:
I believe you need to write all the chunks in a zookeeper transaction to truly get this right: 

https://zookeeper.apache.org/doc/r3.4.3/api/org/apache/zookeeper/ZooKeeper.html#transaction()

MichaelDrogalis:
I don't think it's worth going way out of our way to make sure that multiple jobs under the same ID with different structures are submitted. Per the docs that I added, it's undefined behavior to submit two differing jobs with the same ID. I don't have an opinion yet on your argument for UUID-only job IDs, but I don't want to go wild stopping users from doing something that we've documented as a GIGO operation.

lbradstreet:
They don't have to have different structures. They could just have slightly different data in say the windows data (e.g. the current time) to what is in say the task map. If you have a job queue that builds a job twice I could see why it might have different parameters in it if it had to retry a failed job submission. I could definitely see this happening when you build some sophisticated products on top of onyx. I'd be happy to create a new issue for this and close the current one, and circle back to it later.

As for the IDs, if you don't use UUIDs, then you may have two different users sharing a tenancy submit a job with a certain name and both get a successful return codes even though only one actually succeeded. I'd rather avoid these classes of issues completely by sticking with UUIDs which avoid this problem. It'll also make it easier to enforce a data model on any product that builds on top of onyx because job ids are of known type.

MichaelDrogalis:
@lbradstreet This is ready.

lbradstreet:
Should we place the job hash in the metadata? 

I think eventually we'll want to hash all the individual elements and place them in one chunk. This will make it easier to retrieve everything from s3/HDFS/whatever. This can all wait for now though.

MichaelDrogalis:
All 3 are good changes to me.

src/onyx/api.clj
```diff
@@ -167,34 +165,61 @@
...
-          :job-id id
-          :task-ids (zipmap (map :name tasks) tasks)})
+             client (component/start (system/onyx-client peer-client-config monitoring-config))
+             status (extensions/write-chunk (:log client) :job-hash job-hash id)]
+         (if status
+           (serialize-job-to-zookeeper client id job tasks entry)
+           (let [written-hash (extensions/read-chunk (:log client) :job-hash id)]
+             (if (= written-hash job-hash)
+               (serialize-job-to-zookeeper client id job tasks entry)
+               {:success? false}))))
```
lbradstreet: I think all of the submitters should still go through the process of writing all of the job chunks, in case one dies midway. Otherwise we don't know if the submission was successful.

We no longer need to check the return codes of any of the writes. As long as one won, all is well.

lbradstreet: Oh that's what you were already doing.

lbradstreet:
I squashed the whole branch into a single commit to clean up the history and to make a merge with the abs work easier in the future. Closing now.

PR #600, opened 2016-06-17
Zookeeper thread leak, fixes #566

### LINKED ISSUE #566: Onyx leaks threads when watching ZooKeeper
We've found a thread leak in the following line: https://github.com/onyx-platform/onyx/blob/0.9.x/src/onyx/log/zookeeper.clj#L255

MichaelDrogalis:
Fixes #566 by using a kill channel to unblock the subscriber when the ZK connection closes.


lbradstreet:
Looks good. I'm OK to merge it. I think I will want to quickly check the thread counts from jepsen flight recorder dumps when I get a chance to run jepsen before release.

PR #601, opened 2016-06-17
Only write out one exhausted-input per peer, fixes #493

### LINKED ISSUE #493: Inputs should not write more :exhaust-input log entries than necessary
Seen while Jepsening. A peer will write an exhaust-input entry once per drained? check per task-lifecycle. This could get pretty dangerous when batch-timeout is low. It should be easy to check whether one has already been registered in :exhausted-inputs, and thus not write additional entries out.

MichaelDrogalis:
Fixes #493. As per the issue, I don't think it's a good idea to check the local replica if an exhausted input has been seen because in between the time that the log entry is written and until its been integrated back into the local replica, this entry will continue to be written. With a low batch timeout this could still cause some damage.

Instead, I've added a simple flag to only write to the outgoing channel exactly once.

### DIFF
--- a/src/onyx/information_model.cljc
+++ b/src/onyx/information_model.cljc
@@ -1480,7 +1480,8 @@ may be added by the user as the context is associated to throughout the task pip
                :onyx.core/compiled
                :onyx.core/drained-back-off 
                :onyx.core/messenger-buffer 
-               :onyx.core/state]
+               :onyx.core/state
+               :onyx.core/emitted-exhausted?]
    :env-config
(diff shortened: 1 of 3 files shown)


lbradstreet:
This is fine for now, but I'm not sure it'll last through ABS. We may want to get rid of the exhaust input log entries at that point anyway - rewinding the stream will make it so that we're essentially setting things back into an unexhausted state.

MichaelDrogalis:
Yup, agreed. Definitely fine with axing the whole concept when it's possible.

Issue #471, opened 2016-01-06
Bookkeeper fails to start if cookie already exists in zookeeperBookkeeper fails to start if cookie already exists in zookeeper:


stefanbacon:
Bookkeeper fails to start if cookie already exists in zookeeper:

```
Caused by: org.apache.bookkeeper.bookie.BookieException$InvalidCookieException: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /onyx/test/ledgers/cookies/xxx.xxx.xxx.xxx:3196
    at org.apache.bookkeeper.bookie.Bookie.checkEnvironment(Bookie.java:366)
    at org.apache.bookkeeper.bookie.Bookie.<init>(Bookie.java:466)
    at org.apache.bookkeeper.proto.BookieServer.newBookie(BookieServer.java:107)
    at org.apache.bookkeeper.proto.BookieServer.<init>(BookieServer.java:95)
    at org.apache.bookkeeper.proto.BookieServer.<init>(BookieServer.java:87)
    at onyx.state.bookkeeper.Bookie.start(bookkeeper.clj:35)
    at onyx.state.bookkeeper.BookieMonitor.start(bookkeeper.clj:71)
    at onyx.state.bookkeeper.BookieServers$fn__28021.invoke(bookkeeper.clj:105)
    at clojure.core$mapv$fn__6727.invoke(core.clj:6616)
    at clojure.lang.PersistentVector.reduce(PersistentVector.java:333)
    at clojure.core$reduce.invoke(core.clj:6518)
    at clojure.core$mapv.invoke(core.clj:6616)
    at onyx.state.bookkeeper.BookieServers.start(bookkeeper.clj:104)
    at com.stuartsierra.component$fn__16299$G__16293__16301.invoke(component.clj:4)
    at com.stuartsierra.component$fn__16299$G__16292__16304.invoke(component.clj:4)
    at clojure.lang.Var.invoke(Var.java:379)
    at clojure.lang.AFn.applyToHelper(AFn.java:154)
    at clojure.lang.Var.applyTo(Var.java:700)
    at clojure.core$apply.invoke(core.clj:632)
    at com.stuartsierra.component$try_action.invoke(component.clj:116)
    at com.stuartsierra.component$update_system$fn__16350.invoke(component.clj:138)
    at clojure.lang.ArraySeq.reduce(ArraySeq.java:114)
    at clojure.core$reduce.invoke(core.clj:6518)
[...]

PR #604, opened 2016-06-22
Bk cookie fix

gardnervickers:
This should take care of #471. I expected there to be a way to do this through the API's provided by BookKeeper but it seems that the only Cookie updating/deletion methods are wrapped up in their CLI administration tools, based off of private classes.


src/onyx/state/bookkeeper.clj
```diff
@@ -43,23 +40,34 @@
...
                         (.setDiskUsageThreshold disk-usage-threshold)
                         (.setDiskUsageWarnThreshold disk-usage-warn-threshold))
-          server (BookieServer. server-conf)
-          _ (info "Starting BookKeeper server on port" port)
-          _ (.start server)]
-      (when (:onyx.bookkeeper/delete-server-data? env-config) 
-        (.addShutdownHook (Runtime/getRuntime) 
+          server (try (BookieServer. server-conf)
+                      (catch Exception e
+                        (if (instance? KeeperException$NodeExistsException (.getCause e))
```
MichaelDrogalis: You can just do `(catch KeeperException$NodeExistsException` instead.

lbradstreet:
Merged into develop manually. Thanks.

PR #610, opened 2016-07-04
Catch invalid cookie exception when starting Bookie

jcf:
**Research** around invalid cookie exception when starting an embedded test system in https://github.com/onyx-platform/onyx-kafka/pull/26.

/cc @gardnervickers 

```
16-Jul-04 18:52:48 mbp.local INFO [onyx.static.logging-configuration] - Starting Logging Configuration
16-Jul-04 18:52:48 mbp.local INFO [onyx.log.zookeeper] - Starting ZooKeeper server
16-Jul-04 18:52:49 mbp.local INFO [onyx.state.bookkeeper] - Deleting existing Bookie cookie /onyx/testcluster/ledgers/cookies/192.168.1.51:3196
16-Jul-04 18:52:49 mbp.local FATAL [onyx.system] -
                                   clojure.main.main          main.java:   37
                                                 ...
                                   clojure.main/main           main.clj:  384
                                   clojure.main/main           main.clj:  421
                               clojure.main/null-opt           main.clj:  342
                             clojure.main/initialize           main.clj:  308
                               clojure.main/init-opt           main.clj:  277
                            clojure.main/load-script           main.clj:  275
                                                 ...
                                         user/eval91          REPL Input
                                      user/eval91/fn          REPL Input
                                   user/eval91/fn/fn          REPL Input      (repeats 2 times)
                                  clojure.core/apply           core.clj:  646
                                                 ...
                              clojure.test/run-tests           test.clj:  767 (repeats 2 times)
                                  clojure.core/apply           core.clj:  648
                                                 ...
                                 clojure.core/map/fn           core.clj: 2646
                                clojure.test/test-ns           test.clj:  757
                          clojure.test/test-all-vars           test.clj:  736
                              clojure.test/test-vars           test.clj:  730
                        clojure.test/default-fixture           test.clj:  686
                           clojure.test/test-vars/fn           test.clj:  734
                        clojure.test/default-fixture           test.clj:  686
                        clojure.test/test-vars/fn/fn           test.clj:  734
                               clojure.test/test-var           test.clj:  716
                            clojure.test/test-var/fn           test.clj:  716
                          onyx.plugin.output-test/fn    output_test.clj:   57
                  onyx.test-helper.OnyxTestEnv/start    test_helper.clj:  135
                      onyx.test-helper/try-start-env    test_helper.clj:   89
                                  onyx.api/start-env            api.clj:  346
                                  onyx.api/start-env            api.clj:  349
[...]

PR #613, opened 2016-07-05
Bookie in-memory server cleanup on shutdown

gardnervickers:
Previously we were catching an exception and deleting a node in Zookeeper, what we really want to do is provide the same functionality as the Bookkeeper Admin tool, where you can [format](https://bookkeeper.apache.org/docs/r4.3.0/apidocs/org/apache/bookkeeper/client/BookKeeperAdmin.html#format(org.apache.bookkeeper.conf.ClientConfiguration, boolean, boolean)) the local filesystem and ZooKeeper filesystem. 

We trigger this when `:onyx.bookkeeper/delete-server-data? true`, if we detect there's a mismatched cookie/directory, then we should inform the user that `:onyx.bookkeeper/delete-server-data? true` can be used. There are other ways to recover BookKeeper data and we don't want to destroy the Cookie without the user explicitly instructing us to.

This is in relation to and replaces #610

### DIFF
--- a/src/onyx/state/bookkeeper.clj
+++ b/src/onyx/state/bookkeeper.clj
@@ -73,7 +73,7 @@
           (throw (ex-info "The Bookie server failed to start because a cookie or
                            ledger already exists for this host. Set
                            :onyx.bookkeeper/delete-server-data? true to format
-                           the Bookie environment."))
+                           the Bookie environment on component shutdown."))
           (throw e)))))
   (stop [{:keys [server server-conf] :as component}]
     (info "Stopping BookKeeper server:")
--- a/src/onyx/state/bookkeeper.clj
+++ b/src/onyx/state/bookkeeper.clj
@@ -56,6 +56,10 @@
                           (.setDiskUsageThreshold disk-usage-threshold)
                           (.setDiskUsageWarnThreshold disk-usage-warn-threshold))
             server (BookieServer. server-conf)]
+        (when (:onyx.bookkeeper/delete-server-data? env-config)
+          (.addShutdownHook (Runtime/getRuntime)
+                            (Thread. (fn []
+                                       (format-bk-server server-conf (:conn log))))))
         (info "Starting BookKeeper server on port" port)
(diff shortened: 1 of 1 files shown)


src/onyx/state/bookkeeper.clj
```diff
@@ -4,71 +4,80 @@
...
+               :server server
+               :server-conf server-conf
+               :port port))
+      (catch Exception e
+        (if (or (instance? BookieException$InvalidCookieException (.getCause e))
+                (instance? KeeperException$NodeExistsException (.getCause e)))
+          (throw (ex-info "The Bookie server failed to start because a cookie or
+                           ledger already exists for this host. Set
+                           :onyx.bookkeeper/delete-server-data? true to format
+                           the Bookie environment on startup."))
```
lbradstreet: It actually only formats on shutdown.

src/onyx/state/bookkeeper.clj
```diff
@@ -4,71 +4,80 @@
...
-                                                    ledgers-root-path
-                                                    (Bookie/getBookieAddress server-conf))]
-                            (info "Deleting existing Bookie cookie" cookie-path)
-                            (zk/delete (:conn log) cookie-path)
-                            (BookieServer. server-conf))
-                          (throw e))))]
-      (info "Starting BookKeeper server on port" port)
-      (.start ^BookieServer server)
-      (when (:onyx.bookkeeper/delete-server-data? env-config)
-        (.addShutdownHook (Runtime/getRuntime)
```
lbradstreet: I think we should keep the shutdown hook. Sometimes components aren't shutdown properly, but we still want to format it.

lbradstreet:
Still needs this fixed up to say "shutdown"
```
"The Bookie server failed to start because a cookie or
+                           ledger already exists for this host. Set
+                           :onyx.bookkeeper/delete-server-data? true to format
+                           the Bookie environment on startup."
```
and reflection checks fixed, then I'm good to merge.

PR #617, opened 2016-07-06
Fix flow conditions where flow/from and flow/to is :all

dignati:


lbradstreet:
Looks good. Thank you!

PR #630, opened 2016-07-29
validate config in feedback-exception method

### LINKED ISSUE #629: Add schema annotation to feedback-exception!
If you're being stupid like me this morning you might pass the wrong thing to [`feedback-exception!`](https://github.com/onyx-platform/onyx/blob/28a1d4298d473e7e393e2bffeea3e2b0a1a7bbde/src/onyx/test_helper.clj#L12-L23) and then get some pretty [gnarly errors](https://gist.github.com/jcf/9fcc0c26add7080bbde440e3f3cfe860).

In my case I mistakenly passed the peers rather than the peer config and was told I was not supposed to pass a vector to `onyx.log.zookeeper.Zookeeper/start` thanks to the use of `validate` in that component.

If we can add a schema annotation to `feedback-exception!` to make sure a peer config is passed rather than say the peers themselves that'd be great!

mariusz-jachimowicz-83:
Fixes #629

### DIFF
--- a/src/onyx/system.clj
+++ b/src/onyx/system.clj
@@ -138,12 +138,12 @@
      :log (component/using (zookeeper peer-config) [:monitoring :logging-config])})))
 
 (defn onyx-client
-  ([peer-config]
-   (onyx-client peer-config {:monitoring :no-op}))
-  ([peer-config monitoring-config]
+  ([peer-client-config]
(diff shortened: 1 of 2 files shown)


MichaelDrogalis:
Thanks @mariusz-jachimowicz-83! Merging. :)

PR #634, opened 2016-08-05
Move monitoring config into peer config

dignati:


MichaelDrogalis:
Cheers @dignati. Will review this as soon as I can.

@lbradstreet What do you think about timing for integrating this patch since it's a backwards breaking change? I don't think it's terrible disruptive, but would like to consider when is best.

dignati:
Sure, this can wait. Let me know if you would like to see some changes, I basically just searched for `monitoring-config` and used the `peer-config` or, for `start-env`, `env-config` instead. Maybe I missed something.

MichaelDrogalis:
I think that should pretty much do it, yep.

lbradstreet:
Sorry, I totally missed this. I'll give this a look tomorrow. Thanks @dignati.

lbradstreet:
Happy with this. Thanks!

Issue #621, opened 2016-07-08
Open an HTTP endpoint for every Peer Group for remote query


MichaelDrogalis:
Every peer group should optionally open an HTTP endpoint for health checks and status reports. This facilitates querying in a pull based manner, which complements the pushed based functionality that we already provide.

dignati:
Exposing what lib-onyx already provides([replica-query](https://github.com/onyx-platform/lib-onyx/blob/master/src/lib_onyx/replica_query.cljc) and [job-query](https://github.com/onyx-platform/lib-onyx/blob/master/src/lib_onyx/job_query.clj)) would probably a good start I think? Or do you have something else in mind?

lbradstreet:
Yep, that's essentially what I'm thinking.

The main problem is that it won't be able to use the lib-onyx subscriber because the peer group manages a replica and need to perform transitions, and we don't want a second log subscriber. This should be relatively easy to work around though, we just need the peer group to update the replica atom used by the HTTP server end point.

It'd be good to have an option to turn it on and off via the peer-config.

lbradstreet:
Done

PR #637, opened 2016-08-17
Peer group http endpoint

dignati:
I wanted to get initial feedback on the direction and approach I'm taking in implementing #621.
So far I have copied the http endpoints for replica queries from lib-onyx and made the peer-group-manager update a replica atom on `:apply-log-entry`


MichaelDrogalis:
Oh wow, I didn't expect anyone to work on this. Fantastic, we really wanted to use this in the near term, and this wasn't a small amount of work. Nice job @dignati :)

cc @lbradstreet

lbradstreet:
This is awesome! Well done with the implementation - everything looks great. Now that it's actually done, my only concern is bloating the dependencies for users that want a version without the end point.

One option is to split the HTTP server component out into a library, and include it by default. This would allow users to easily exclude the dependency, and turn it off in the peer config.

What do you think?

dignati:
Making the server an optional dependency sounds good to me. Maybe it's best to require the namespaces on demand when ` onyx.query/server?` is set and have the user require the needed libraries in their project? I think this is how datomic handles requirements for different storage backends.

lbradstreet:
A good approach that we've used before is to convert the function that builds (here https://github.com/onyx-platform/onyx/pull/637/files#diff-fdca433249c240a7b884e4d280b868b7R153) it to a multimethod that dispatches on the type of server, or maybe whether the server? setting is true or false. Then set the default case to throw an exception, saying that if you want to use the server you need to add the dependency and require X namespace.

dignati:
Ah yes, the multimethod approach sounds like a good fit. The user still has to require the ring and jetty dependencies in their own project.clj then, right?

dignati:
Also, should I try to import the job queries from lib-onyx in this PR?

lbradstreet:
What I was thinking is that we could split the http server that you've incorporated into onyx core into its own project/library. This library depends on cheshire, ring, etc. Then any user wishing to add a http replica server adds this library as a dependency and requires the namespace that has the defmethod that builds the server record.

I would prefer to promote the web server from lib-onyx to its own project.

Hit me up on slack if you want clarification on anything.

lbradstreet:
Merging. Will split the rest of the code out into its own project after merge.

lbradstreet:
Query code moved into https://github.com/onyx-platform/onyx-node-query. Needs proper README, and to be part of the release process

MichaelDrogalis:
Also needs docs. I will try to take care of the above and this one this weekend.

PR #638, opened 2016-08-18
WIP: filter using lmdb

### LINKED ISSUE #583: Switch from RocksDB to LMDB
I'm convinced this will be better/faster/easier for us.

mariusz-jachimowicz-83:
Resolve #583 
- [x] filter implementation
- [x] add tests


lbradstreet:
This PR is going to be on the back burner for a while. The work will definitely get used in the future - one way or another, but I wouldn't put any more work into it for now. Thanks!

lbradstreet:
The filters are no longer used in ABS, but I will likely re-use some of this code in the future to back our window management code. Thanks!

PR #646, opened 2016-09-08
Fix typos and broken links.

codonnell:
I don't know how you build your docs, so I haven't tried compiling with these changes. I believe they should work, but definitely double check.


MichaelDrogalis:
Thanks!

PR #649, opened 2016-09-20
Support :onyx/fn batches which return segments

lbradstreet:
smw from clojurians will bring this home (need to find out his github name)

I've added a test and added it to the information model and the schema. It needs to be tested by smw to make sure it works with his use case, and hammer on it a little more. See https://github.com/onyx-platform/onyx/blob/d6fdbe4f1b290fe144bb192f48c79245b88622de/test/onyx/peer/batch_function_test.clj for the testing that's been done so far.

It also needs schema added so that users can't use `:onyx/bulk?` at the same time as `:onyx/batch?`. Please see this section of schema.cljc to see how it can be done https://github.com/onyx-platform/onyx/blob/d6fdbe4f1b290fe144bb192f48c79245b88622de/src/onyx/schema.cljc#L221.

In hindsight we would probably make `:onyx/bulk?` take a keyword option that chooses modes. One option is to keep supporting `:onyx/bulk?` but make `:onyx/batch` take an option like `:side-effects` or `:segments`. `:onyx/bulk?` would then turn `:onyx/batch :side-effects` on.


lbradstreet:
@smw @MichaelDrogalis, I've modified this to take `:onyx/batch-fn?` and deprecated `:onyx/bulk?` with a good explanation of how to transition between them. I also loosened up the requirement in the doc that the task be a function task, as I can see it being useful on other task types. I also spell checked the information model doc and found some non batch mistakes.

@smw could you please test that this works for your purposes.

Issue #627, opened 2016-07-22
Exceptions written via kill-job should have a fallback string version


lbradstreet:
Sometimes exceptions written by kill job will contain types that are not deserializable by a JVM that did not produce them. We should either write a string version of the exception, or provide a better error in feedback-exception! / onyx-dashboard when the exception can't be deserialised

lbradstreet:
Fixed thanks to  #653. Thanks @colinhicks!

PR #653, opened 2016-09-26
Normalize lifecycle exceptions for deserialize-ability

colinhicks:
Aims to resolve onyx-platform/onyx#627

From @lbradstreet:
> Sometimes exceptions written by kill job will contain types that are not deserializable by a JVM that did not produce them. We should either write a string version of the exception, or provide a better error in feedback-exception! / onyx-dashboard when the exception can't be deserialised

My (newbie) assumptions:
* The relevant exceptions are written to the log when some part of the task lifecycle throws.
* We could see the deserialization issue, for example, if a peer (JVM A), throws an exception that is read off the log by a consuming client like onyx-dashboard (JVM B), where the classpath does not include the exception type.

This PR's approach wraps the original throwable in an ex-info that preserves the exception message and stack trace. The ex-data key `:original-exception` carries the keyword form of the original exception class name.

Compared to the string-version approach, I think this has the advantage of maintaining a Throwable instance as the log entry's :exception value. And it has less surface area compared to handling the issue case downstream in the dashboard or tests. Tradeoff: consuming code can't catch specific exception types; `:original-exception` would need to be leveraged instead.

Thoughts? Hope this helps!

### DIFF
--- a/src/onyx/compression/nippy.clj
+++ b/src/onyx/compression/nippy.clj
@@ -9,9 +9,29 @@
                                                   :str (.getMessage x)
                                                   :cause (.getCause x)})))
 
+(defn exception-info-reader [tag value]
+  (let [maybe-record? (pos? (.indexOf (name tag) "."))]
+    (if-not maybe-record?
+      {:tag (keyword tag)
+       :value value
+       :deserialize-attempted? false}
+      (try
+        (let [rec-class (Class/forName (name tag))
+              create-method (.getMethod rec-class
+                                        "create"
+                                        (into-array Class [clojure.lang.IPersistentMap]))]
+          (.invoke create-method rec-class (into-array Object [value])))
+        (catch Throwable ex
+          {:tag (keyword tag)
+           :value value
+           :deserialize-attempted? true
+           :deserialize-exception-type (keyword (type ex))})))))
+
 (nippy/extend-thaw :onyx/exception-info
                    [data-input]
-                   (let [{:keys [ex str cause]} (read-string (.readUTF data-input))]
+                   (let [{:keys [ex str cause]}
+                         (clojure.edn/read-string {:default exception-info-reader}
+                                                  (.readUTF data-input))]
                      (clojure.lang.ExceptionInfo. str ex cause)))
 
 (def messaging-compress-opts {})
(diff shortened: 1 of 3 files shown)


lbradstreet:
Thanks! This is much better than what I was suggesting. Unfortunately, it suffers from the same problem as the original. Certain types of objects may be contained in the ex-info data (or otherwise included in the exception) may not be deserializable, especially on the dashboard. This is an especially hard problem because we cannot rely on nippy's judgement about whether the objects are properly serializable, as it may be serializable on the onyx peer which has imported classes that the dashboard will not have (see https://github.com/ptaoussanis/nippy/blob/master/src/taoensso/nippy.clj#L895, note that this is a later version that what we use).

I don't have a good way forward for you yet. I'll have to think about it a little more. The best would be to transform the ex-data in a particular way, without being too lossy for clojure's value types, but this could be tricky if nippy isn't helping us forward.

colinhicks:
Thanks for your response. Now that I understand the possibly unresolvable types in ex-info aspect, I took a pass at altering the way Nippy thaws ExceptionInfos in 6900a0f.

Example deserialized ex-data:
```clojure
{:consumer
 ;; The custom reader does not re-instantiate objects ...
 {:tag :object,
  :value [some.ns.FranzConsumer 791519068 "some.ns.FranzConsumer@2f2d9f5c"],
  :deserialize-attempted? false},
 :event
 ;; But it does attempt to recreate record instances.
 ;; This one was resolved, so we get a normal record instance.
  #onyx.types.Event{:monitoring nil, :metrics nil}
 :thing
 ;; This record was unavailable in the consuming context.
 ;; But the ex-data preserves its type and value.
 {:tag :some.other-ns.ThingRecord,
  :value {:id 1234},
  :deserialize-attempted? true,
  :deserialize-exception-type java.lang.ClassNotFoundException}}
```

Totally understand if this is still off on the wrong track. Happy to stand down on this PR if this isn't suitable or if it's adding to churn.

lbradstreet:
Great! This is looking better. I'll need to have a play with it before merging, but it's looking good so far.

Thanks

lbradstreet:
After some testing, this looks good, and will be released as part of 0.9.11. Thanks for your contribution!

PR #654, opened 2016-09-28
Log tenancy-id with zero-peers message

### LINKED ISSUE #652: Add :onyx/tenancy-id to "no virtual peers" error when submitting a job
See the following example log message that you'll see when a job can't be started:

```
2016-09-22T17:37:49.810 [async-thread-macro-111] INFO  onyx.log.commands.submit-job - Job ID ee99b87c-1484-4d10-8ac7-39a49b0df978 has been submitted, but received no virtual peers to start its execution.
                     Tasks each require at least one peer to be started, and may require more if :onyx/n-peers or :onyx/min-peers is set.
                     If you were expecting your job to start now, either start more virtual peers or choose a different job scheduler, or wait for existing jobs to finish.
```

I was getting this message maybe 100 times, and couldn't work out why because I was only submitting a single job.

When I mentioned this to @lbradstreet in Gitter he suggested straightaway that I was probably using the same tenancy ID over and over when I should be using my generated one.

If we include the tenancy ID in the error message it may be obvious to others that something's not right in his or her Onyx configuration.

Thanks again @lbradstreet!

colinhicks:
Resolves #652.

### DIFF
--- a/src/onyx/log/commands/kill_job.clj
+++ b/src/onyx/log/commands/kill_job.clj
@@ -18,6 +18,8 @@
           (update-in [:killed-jobs] conj job-id)
           (update-in [:killed-jobs] vec)
           (update-in [:allocations] dissoc job-id)
+          (update-in [:exhausted-inputs] dissoc job-id)
+          (update-in [:sealed-outputs] dissoc job-id)
           (update-in [:task-metadata] dissoc job-id)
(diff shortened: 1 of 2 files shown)


MichaelDrogalis:
Hooray! Thank you! :)

PR #672, opened 2016-11-12
Balanced scheduler allocations

lbradstreet:
This PR fixes a bug where jobs allocated by the balanced job scheduler potentially act in a greedy way once they are already allocated. This PR ensures that peers are properly spread over coverable jobs.

An additional, unrelated, workaround is included for acking / completing segments on non-input tasks after they are reallocated. This is not the best solution as this mechanism will be replaced soon.


lbradstreet:
Appears to be fixed

PR #673, opened 2016-11-15
Document change regarding flux-policy

yonatane:
Stating there are three flux policies instead of two + fixing a minor typo


MichaelDrogalis:
Thank you!

MichaelDrogalis:
(Build failed due to network issue, for the record. This patch is fine.)

PR #685, opened 2016-11-23
Ensure test env components always stop

ChrisBlom:
I had some problems with the test environment: some of its components were not stopped 
if other components failed to start, as a result ports were not freed and I needed to restart the repl.

For such cases I have this `with-components` macro lying around, which ensures that all components stop, even if starting some components throws exceptions. 
In this PR I used it to ensure the test environment components are always stopped.

The test environment is now now longer a Record that implements Lifecycle, but just a map,
which has the same fields as before.


MichaelDrogalis:
I'm good with this patch conceptually. I've had a few snags now and again with ports not being freed, it's terribly annoying. Will take a close look when the tests pass.

ChrisBlom:
Ok, tests are passing now

MichaelDrogalis:
I'm fine with this patch. @lbradstreet, thoughts?

lbradstreet:
Yes, this looks good to me. Merged.

lbradstreet:
This is much better, but I've still managed to ctrl-c it at an inopportune time and have it keep the ZooKeeper ports bound. I will create an issue to track any remaining issues.

Thanks again.

PR #729, opened 2017-01-20
Cljs spec compatability

bamarco:
Addresses https://github.com/onyx-platform/onyx-local-rt/issues/6 

If you change a few ns parameters then the recent cljs compiler won't choke on spec.cljc.


MichaelDrogalis:
@bamarco Thanks! Looks great. Merging.

PR #738, opened 2017-02-20
Allow specifying `:onyx.peer/storage.s3.endpoint`

gardnervickers:


gardnervickers:
Created entries for the schema and information model.

PR #743, opened 2017-03-17
Add (:gen-class) directive to allow

gardnervickers:
dependent projects to use the media driver
as a jar entrypoint.


lbradstreet:
Great. Thanks!

PR #745, opened 2017-03-25
add couple of missing =

AndreaCrotti:
I just noticed a problem in the README, simply missing a couple of equals in the img definition.


MichaelDrogalis:
Strange, I wonder how that happened. Thanks!

PR #747, opened 2017-03-27
Dynamic write-batch size and new serializer

lbradstreet:


src/onyx/messaging/aeron/subscriber.clj
```diff
@@ -75,6 +82,7 @@
     (let [error-handler (reify ErrorHandler
                           (onError [this x] 
                             (reset! error x)
+                            (info "error" x)
```
MichaelDrogalis: Switch to an error Timbre call.

src/onyx/plugin/messaging_output.clj
```diff
@@ -61,9 +27,37 @@
...
+        enc (pub/segment-encoder pub)]
+    (if (zero? (segment-encoder/segment-count enc))
+      (.position ^io.aeron.Publication (.publication pub))
+      (let [segments-len (- (segment-encoder/offset enc)
+                            (base-encoder/length base-enc))] 
+        (base-encoder/set-payload-length base-enc segments-len)
+        (pub/offer! pub (.buffer base-enc) (segment-encoder/offset enc) epoch)))))
+
+(defn try-sends [pubs epoch]
+  (loop [pubs pubs]
```
MichaelDrogalis: Is this a possible hanging point if we encounter an unclean shutdown?

src/onyx/messaging/aeron/subscriber.clj
lbradstreet: This line should be removed.

src/onyx/plugin/messaging_output.clj
lbradstreet: Nope, it'll stop recurring if there is a negative return code.

PR #748, opened 2017-03-28
Fixed readme.md companies images.

jstrength:
They were missing an ='s in their height attribute.


MichaelDrogalis:
Just got this one patched in develop yesterday, we'll be moving it through with a release today. Thanks though!

PR #762, opened 2017-04-30
Small trigger doc improvement

solatis:
It wasn't completely obvious to me what exactly "downstream" meant in the context of triggers, this patch improves that a bit.


lbradstreet:
Thanks!

PR #783, opened 2017-05-17
Publication image timeout safety, general performance work, short circuiting, egress write batch sizing.

lbradstreet:
Also implements ticket GCing when new replica versions invalidate old tickets.


src/onyx/flow_conditions/fc_routing.cljc
```diff
@@ -56,19 +56,18 @@
    (->Route #{} #{} nil nil nil)
    compiled-flow-conditions))
 
-(defn route-data [{:keys [egress-tasks compiled-ex-fcs compiled-norm-fcs onyx.core/task-id
```
lbradstreet: General perf work, only look things up when needed.

PR #800, opened 2017-09-05
Windowing improvements

lbradstreet:


src/onyx/peer/task_lifecycle.clj
```diff
@@ -203,7 +203,7 @@
           metadata-bs (checkpoint-compress {:state-indexes (ws/state-indices event)})
           _ (cpenc/set-schema-version checkpoint-encoder 0)
           _ (cpenc/set-metadata checkpoint-encoder metadata-bs)
-          exported-state (db/export (get-state-store state) checkpoint-encoder) 
+          _ (db/export (get-state-store state) checkpoint-encoder)
```
MichaelDrogalis: Intentional/necessary? Is this side-effecting?

lbradstreet: Yes, it serializes it into the encoder, so the old binding was incorrect.

lbradstreet:
One last thing to do before merge is mention that the aggregate create-state-update signature has changed.

PR #805, opened 2017-10-04
Native watermarks support.

lbradstreet:
Implements stream level watermarks at barrier level granularity.
Limitations:

Depends on the barriers to transfer the watermarks down. If your checkpointing period is high, then you will only process the watermarks with that granularity. Future work will implement watermark messages which have higher granularity.
If you use trigger/emit and flush a lot of segments from a window downstream, you will see a corresponding memory increase. Future work should page these segments in incrementally.
Makes a breaking change to plugin interface. We should probably add an extra plugin interface (Watermarkable?) and then check whether the plugin implements it. That would prevent this from being a breaking change.


src/onyx/information_model.cljc
```diff
@@ -304,14 +304,21 @@
...
                    :tags [:function]
                    :required-when ["`:onyx/type` is set to `:function`"]
                    :optionally-allowed-when ["`:onyx/type` is set to `:input`"
                                              "`:onyx/type` is set to `:output`"]
                    :added "0.8.0"}
 
+                  :onyx/assign-watermark-fn
+                  {:doc "A function to assign a watermark to a datasource by inspecting a segment read from that datasource. Should return the numbers of milliseconds since epoch. Missing watermarks will be ignored. A fully qualified, namespaced keyword that points to a function on the classpath. "
+                   :type :keyword
+                   :tags [:function]
```
lbradstreet: tags :input, tags :windowing?

src/onyx/information_model.cljc
```diff
@@ -684,6 +691,13 @@
              :optional? true
              :added "0.8.0"}
 
+            :trigger/delay
+            {:doc "Used with the trigger `:onyx.triggers/watermark`. A watermark trigger applies after the watermark has passed the end of a window. `:trigger/delay` adds an additional delay to the watermark, such that a window will only be triggerd after the watermark + delay."
```
lbradstreet: triggerd.

lbradstreet: Include an example.

src/onyx/information_model.cljc
```diff
@@ -840,6 +854,11 @@ may be added by the user as the context is associated to throughout the task pip
              :type :integer
              :optional? true
              :added "0.9.0"}
+            :watermarks
+            {:doc "Job level watermark times, effective at this peer on this task."
```
lbradstreet: Includes a map looks like `{:input watermarked-time-since-epoch :coordinator times-since-epoch}`

src/onyx/peer/task_compile.clj
```diff
@@ -11,8 +11,16 @@
...
             [onyx.static.logging :as logging]
             [onyx.refinements]
+            [onyx.peer.operation :refer [resolve-fn]]
             [onyx.windowing.window-compile :as wc]))
 
+(defn resolve-watermark-fn [task-map]
+  (if-let [fkw (:onyx/assign-watermark-fn task-map)]
+    (let [f (resolve-fn {:onyx/fn fkw})] 
+      (fn [segment] 
+        (or (try (f segment) (catch Exception _))
```
lbradstreet: Error metric on this one?

src/onyx/triggers.cljc
```diff
@@ -115,11 +102,11 @@
...
 (defn watermark-fire?
-  [trigger trigger-state {:keys [upper-bound event-type segment window] :as state-event}]
-  ;; If this was stimulated by a new segment, check if it should fire.
-  ;; Otherwise if this was a completed task, always fire.
-  (or (and segment (exceeds-watermark? window upper-bound segment))
-      (#{:job-completed :recovered} event-type)))
+  [{:keys [trigger/period]} trigger-state {:keys [event-type upper-bound watermarks] :as state-event}]
+  (or (= :job-completed event-type) 
+      (and (= :watermark event-type)
+           (let [standard (if period (apply to-standard-units period) 0)]
```
lbradstreet: rename that to delay

src/onyx/triggers.cljc
```diff
@@ -87,11 +79,6 @@
   (let [{:keys [pred-fn]} trigger-state]
     {:fire? (pred-fn trigger state-event)}))
 
-(defn watermark-next-state
-  [trigger state state-event]
-  ;; Intentionally return nil - this trigger is stateless.
-  )
-
 (defn percentile-watermark-next-state
   [trigger state state-event]
```
lbradstreet: switch over to extent stateful trigger.

src/onyx/triggers.cljc
lbradstreet: >=

test/onyx/peer/resume_points_test.clj
```diff
@@ -33,6 +33,7 @@
 (defn update-atom! [event window trigger 
                     {:keys [lower-bound upper-bound event-type] :as opts} 
                     extent-state]
+  (println "COMPLETED" event-type)
```
lbradstreet: delete

test/onyx/triggers/trigger_watermark_test.clj
```diff
@@ -4,43 +4,43 @@
...
-                 :trigger/id :trigger-id}
-        segment {:event-time t}
-        event {}]
-    (is
-     (not
-      (onyx.triggers/watermark-fire? trigger nil 
-                                     (map->StateEvent {:window window 
-                                                       :segment segment
-                                                       :upper-bound (inc t)}))))))
+;                  :trigger/on :onyx.triggers/watermark
```
lbradstreet: delete this test.

PR #818, opened 2017-10-18
Alter handling of trigger/emit'd segments

lbradstreet:
by placing them in onyx.core/results rather than onyx.core/triggered.
This allows them to be automatically processed by output plugins.


lbradstreet:
Ensure the trigger/emit order is correct in onyx.core/results before merge.

lbradstreet:
Pulled in to master via other branch.

PR #825, opened 2017-11-08
Disallow ZK windowing by default.

MichaelDrogalis:


lbradstreet:
Update the doc to say: "Allows window contents to be checkpointed with ZooKeeper. This is highly unadvised for anything but testing, as ZooKeeper checkpoints are not written asynchronously, are not automatically garbage collected, and ZooKeeper does not support znodes greater than 1MB."

MichaelDrogalis:
Done.

lbradstreet:
Merged!

PR #826, opened 2017-11-10
Checkpoint GC

MichaelDrogalis:


lbradstreet:
Made changes:
- no longer delete final watermark epoch znode, future GCs.
- allow tenancy to be supplied to GC non-current tenancies.
- added (bad) test
- added onyx.api/clear-checkpoints to clear all old checkpoints.

Ready for re-review @MichaelDrogalis.

PR #832, opened 2017-12-10
Make 0.10.x compatible with clojure 1.9

souenzzo:
`clojure-futre-spec` is compatible with clojure 1.9¹, but when refer a specific function to import, clojure 1.9 fail with `IllegalAccessError uuid? does not exist`

¹ https://github.com/tonsky/clojure-future-spec#clojure-future-spec


lbradstreet:
Hi @souenzzo. We have split out clojure future spec in 0.12.x, and 0.10.x is no longer supported. We strongly suggest you upgrade or release your own dependency until you get a chance to upgrade.

Thanks anyway!

PR #836, opened 2017-12-31
Update aggregation state management docs

twashing:
Updating the code samples to correspond with the 0.12.x API. Specifically, the `:aggregation/create-state-update` fn signature has changed from `(window, state, segment)` to `(window, segment)` (see [here](https://github.com/onyx-platform/onyx/blob/0.12.x/changes.md#0110)).


lbradstreet:
Looks good. Thanks!

PR #837, opened 2018-01-04
no-op-node for task-tagged-constraints

Thingographist:
Balancer for tagged tasks with _n-peer_ constraints


lbradstreet:
Great work. Do tagged tasks now work in your code again?

Thingographist:
Yes. This solution make full model for btrplace and plan for allocations. In reworked tests only change order of allocated nodes.

lbradstreet:
Thank you! I will merge it an include it in 0.12.3.

PR #843, opened 2018-01-30
Add :onyx.peer/storage.s3.protocol configuration option

niamu:
We are using [Ceph](http://docs.ceph.com/docs/master/radosgw/s3/)'s S3 API, but our staging environment isn't served over SSL at the moment.

I think having an option to use HTTP over HTTPS will be a useful configuration option for those in similar situations not using AWS S3.

### DIFF
--- a/src/onyx/information_model.cljc
+++ b/src/onyx/information_model.cljc
@@ -1,14 +1,14 @@
 (ns onyx.information-model)
 
-(def peer-scheduler-event-types 
+(def peer-scheduler-event-types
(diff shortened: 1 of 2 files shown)


niamu:
Sorry about all the whitespace changes. I forgot to disable that in my editor. The changes are rather small once you [disable whitespace in the diff](https://github.com/onyx-platform/onyx/pull/843/files?w=1).

lbradstreet:
Looks good. Thanks!

PR #851, opened 2018-03-09
Support custom checkpoint storage implementations

brianh:
This has been tested with a custom storage implementation and with built-in :zookeeper. I have no easy way of testing it with :s3.

Went ahead and updated the info model so the change would be reflected in the documentation (using some assumptions based on context).  Feel free to change/wordsmith/delete!


MichaelDrogalis:
I'm okay with leaving the info model in tact for the defaults, but the Schema change is required. cc @lbradstreet

MichaelDrogalis:
Looks okay to me. Any final comments?

PR #857, opened 2018-04-28
Remove unused watermark and update documentation for take-segments!

quimrstorres:


src/onyx/plugin/core_async.cljc
```diff
@@ -122,7 +122,6 @@
   (map->AbsCoreAsyncReader {:event event
                             :chan (:core.async/chan event) 
                             :completed? (atom false)
-                            :watermark (atom nil)
```
quimrstorres: @lbradstreet Am I seeing this wrong? The plugin does not make use of this value and it doesn't implement `WatermarkedInput` so I guess this is not used?

PR #858, opened 2018-04-28
Fix broken links

quimrstorres:


README.md
```diff
@@ -118,7 +118,7 @@ Feeling impatient? Hit the ground running ASAP with the [onyx-starter repo](http
 
 ### API Docs 0.12.8-SNAPSHOT
 
-Code level API documentation [can be found here](http://www.onyxplatform.org/docs/api/0.12.8-SNAPSHOT).
```
lbradstreet: This link gets automatically updated when we do a release, and that release gets deployed to onyxplatform.org, so it's fine to keep it. You're right that it's broken for snapshots, but I want to keep it here.

lbradstreet:
We should probably setup a crawler to notify us of broken links on the website, as we do keep generating them over time. If you know of one, I'll give it a go.

README.md
quimrstorres: I think the links for the stable releases are not working as well. You can see that [the link](https://github.com/onyx-platform/onyx#api-docs-0127) in the README for the branch `0.12.x` is also not working, and that points to the 0.12.7 docs. I found about this in that branch. Maybe the docs are not being generated as expected then?

quimrstorres:
Yes, that's a good idea. I'll try to investigate possible tools and get back to you 👍

lbradstreet:
Fixed directly in master as the snapshot fix was fine. Thanks!

PR #859, opened 2018-04-28
Improve error message for flow conditions that are not connected in the workflow

### LINKED ISSUE #855: Flow conditions validation shows bad error message when tasks are not connected.
Task `:in-task` is not connected to `:response-out` in the workflow, and this error message is very bad as it is a valid task name, though it is not connected.
```
------ Onyx Job Error -----
There was a validation error in your flow-conditions for key :flow/to
{
   :flow/predicate :compute.command-processor.tasks.common/response?
   :flow/short-circuit? false
   :flow/from :in-task
   :flow/to [:response-out]
    ^-- [:response-out] isn't a valid task name.
}
Did you mean: :response-out
------
```

quimrstorres:
With this change, the error message looks like this:

```
------ Onyx Job Error -----
There was a validation error in your flow-conditions for key :flow/to
{
   :flow/from :in
   :flow/to [:out]
    ^-- :flow/from and :flow/to must be connected in the workflow.
   :flow/predicate :onyx.validation-test/f
}
-- Docs for key :flow/to --
The destination task where segments will arrive. If set to :all, all
downstream tasks will receive this segment. If set to :none, no downstream
tasks will receive this segment. Otherwise it must name a vector of keywords
indicating downstream tasks. The order of keywords is irrelevant.
Expected type: [:keyword [:keyword]]
Choices: [[:any] :all :none]
Added in Onyx version: 0.8.0
------
```

I guess the footer is not ideal, as this is not necessarily a problem with `:flow/to`, but most of the error machinery is expecting a key to associate the error with.

The previous message can be seen in #855.

Closes #855

### DIFF
--- a/src/onyx/static/helpful_job_errors.cljc
+++ b/src/onyx/static/helpful_job_errors.cljc
@@ -65,7 +65,9 @@
    ["This window type requires a :window/window-key to be defined."]
 
    :auto-short-circuit
-   [":flow/to :all and :none require :flow/short-circuit? to be true."]})
+   [":flow/to :all and :none require :flow/short-circuit? to be true."]
+   :disconnected-tasks
+   [":flow/from and :flow/to must be connected in the workflow."]})
 
 (def relevant-key
   {'task-name? :onyx/name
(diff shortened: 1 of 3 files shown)


lbradstreet:
This is much better! Maybe we could add to the doc that is used in the footer, and mention that flow/from and to expect tasks to be connected in :workflow?

lbradstreet:
Merging. Will add to the error message in master.

lbradstreet:
Thanks again!

PR #860, opened 2018-04-29
Move seq plugin documentation from external repo

quimrstorres:
The seq plugin was moved into `onyx.core` but the documentation hadn't been moved yet. The only changes made to the docs were on how to use the plugin now that it's not an external dependency, changing the format from Markdown to AsciiDoc and small wording tweaks.

The external repo was meanwhile archived but can still be seen here:
https://github.com/onyx-platform/onyx-seq


lbradstreet:
Great! Thanks.

PR #877, opened 2018-10-24
Allow job snapshot coordinates to be looked up by job-name AND tenancy-id

sundbry:
This adds a way for the client to restrict snapshot look-ups by job name to a particular tenancy id.

I was getting checkpoint not found bugs when I had two parallel tenancies running in my zk cluster - then I realized the (job-snapshot-coordinates client job-name) method was using global state and returning jobs with mixed tenancy ids.


lbradstreet:
Looks great. Thanks! I'll cut a release soon.

PR #881, opened 2018-11-10
Fix invalid :import caught with Clojure 1.10 specs

puredanger:
But note that this invalid statement previously had no effect, so you can see the fully-qualified name being used in the source. It might be better to either remove the import or remove the qualification in the source.


lbradstreet:
Thanks Alex! I'll remove the qualification in the source too.

PR #882, opened 2018-11-15
Make onyx.job cljc compatible

thenonameguy:
There is no reason that this file should be Clojure only, we should surface it to as many envs as possible.


lbradstreet:
Cheers!

PR #883, opened 2018-11-15
Catch NoNodeException and return nil for children of non-existing node

sundbry:
NoNodeException was blocking a checkpoint garbage collection when the /epoch/ node did not exist


lbradstreet:
Thanks! Looks good.

PR #889, opened 2019-04-23
Fast method for getting job-state when current-replica is cached

sundbry:
Adds a job-state multimethod which returns the job's state without re-reading from zookeeper, if we already have the current replica information at hand.


lbradstreet:
LGTM! Thanks.

Issue #886, opened 2019-04-18
Onyx hangs when provide ":onyx.messaging.aeron/media-driver-dir" setting in peer-config


darongmean:
I tried to run two onyx instances in the same machine. So I needed to change Aeron media driver directory on both instances. After digging through source code, I found out I could pass the directory via `:onyx.messaging.aeron/media-driver-dir` setting. It was working fine with two onyx instances.

If I run one onyx instance, Onyx hanged with the following message:
`Aeron media driver has not started up. Waiting for media driver before starting peers, and backing off for 500ms.`

I think the problem lying in the following code:
https://github.com/onyx-platform/onyx/blob/0.14.4/src/onyx/peer/peer_group_manager.clj#L264
https://github.com/onyx-platform/onyx/blob/0.14.4/src/onyx/messaging/aeron/messaging_group.clj#L30
The function `media-driver-healthy?` initializes `CommonContext` object with default media driver directory. It fails to consider that the directory might be overridden by `:onyx.messaging.aeron/media-driver-dir` setting.

I could submit a PR. However, I'm new to onyx. I'm not sure this is the right direction. Could anyone help guiding me with this issue?

sundbry:
It sounds like you are on the money with your analysis @darongmean, I would just submit a PR!

PR #891, opened 2019-04-28
Use "media-driver-dir" if provided to determine status of media driver

darongmean:
Address issue #886.


lbradstreet:
Thanks @darongmean. LGTM.

PR #892, opened 2019-04-30
Add optional key `:onyx/percentage` into base-task-map schema.

SuperStevenZ:
Add `:onyx/percentage` as an optional key into base-task-map schema, so that we can set percentage  for each individual catalog entry in job definition when using percentage task scheduler, as suggested [here](http://www.onyxplatform.org/docs/user-guide/0.14.x/#_percentage_task_scheduler).


solatis:
Looks good to me, thanks for the patch!

solatis:
I had to revert this commit due to failing tests. Any chance you could look into this?

SuperStevenZ:
@solatis Yep, I get a bead on it, and will create a new PR to hopefully pass the CI.

Also, thanks for maintaining this project going forward!

PR #895, opened 2019-05-06
Add optional key `:onyx/percentage` into base-task-map schema.

SuperStevenZ:
This is a follow-up PR to fix the failing test in this [PR](https://github.com/onyx-platform/onyx/pull/892).

After adding the information model for `:onyx/percentage`, this [test](https://github.com/onyx-platform/onyx/blob/0.14.x/test/onyx/doc_test.clj#L37-L43) shouldn't fail anymore, and the documents for `:onyx/percentage` should show up in [cheat sheet](http://www.onyxplatform.org/docs/cheat-sheet/latest/#/catalog-entry).

Running `lein test` in my local shows
```shell
$ lein test
...
Ran 160 tests containing 525 assertions.
0 failures, 0 errors.
```

### DIFF
--- a/src/onyx/information_model.cljc
+++ b/src/onyx/information_model.cljc
@@ -219,6 +219,14 @@
                    :optional? true
                    :added "0.8.0"}
 
+                  :onyx/percentage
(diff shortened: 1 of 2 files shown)


solatis:
Thanks, looks good!

PR #897, opened 2019-05-30
Log as :info level when Aeron starts up in a healthy state

sundbry:
The program should not be producing warnings if it starts up normally (which can create false alarms in downstream monitoring)


solatis:
Makes sense, merging!

